PR #97, opened 2014-09-30
Fix RHEL 7 does not have openvpn-auth-ldap package

brunoleon:
At the moment there is no openvpn-auth-ldap package under rhel 7
This fix allow the module to still install, though obviously the ldap feature won't be available


luxflux:
Finally got the time to fix this.

retizu:
I've added some notes here that might be worth considering on the next iteration https://github.com/luxflux/puppet-openvpn/commit/bae0c0e193306c3f03b81d2ed61e710dee4d6a68

luxflux:
There is already an issue for adding `openvpn-auth-ldap`: #172

PR #98, opened 2014-10-01
Rhel7 systemd

brunoleon:
Support RHEL7 systemd.
Also add containment so we can depend on the main class openvpn.


luxflux:
Thanks for your patience. Could you merge master into this branch so that we can start reviewing?

luxflux:
Closing this one as #127 overcomes it.

PR #99, opened 2014-10-01
Log path compatible wih selinux

brunoleon:


luxflux:
As there have been a lot of changes and also some stuff regarding the available package and systemd, could you create a new pull request based on master which brings in the support of "Log path compatible wih selinux"?
It would be cool to fix this issue on SELinux platforms.

PR #116, opened 2015-01-05
Adds in the ability to set rcvbuf and sndbuf parameters.

typhonius:
I use puppet-openvpn.
I read [this article](http://lowendtalk.com/discussion/comment/843561).
I figured it'd be useful to include.
Panda :panda_face:

### DIFF
--- a/manifests/client.pp
+++ b/manifests/client.pp
@@ -101,6 +101,14 @@
 # [*down*]
 #   String,  Script which we want to run when openvpn client is disconneting
 #
+# [*sndbuf*]
+#   Integer, Set the TCP/UDP socket send buffer size.
+#   Default: undef
+#
+# [*rcvbuf*]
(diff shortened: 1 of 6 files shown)


luxflux:
Thanks for your code! :beers: That's really a good thing to add these parameters.

According to the article, the parameters are needed on server and client side. Could you add the client side as well? Thanks a lot!

luxflux:
Ah, could you merge master? I "fixed" the travis build.

typhonius:
- Rebased on upstream/master
- Client config and tests added
- Commits flattened into single commit

Good call on adding client config. I omitted it personally as I cannot change client config on our set up. Instead I push the config:

````
    push         => [
      "sndbuf 393216",
      "rcvbuf 393216",
    ],
    sndbuf       => 393216,
    rcvbuf       => 393216,
````

luxflux:
Thanks a lot!

PR #117, opened 2015-01-08
Don't put the cipher key in if no cipher is specified.

typhonius:
If no cipher is specified in puppet, the client configuration files get written with a cipher key but no value. This causes the android openvpn app to complain.

I envisage 3 possible solutions, of which this PR implements the most simple:
- Copy server.erb and only write the cipher if it not empty
- Default the cipher to a strong cipher ie AES-256-CBC
- Default the cipher to undef so it does not get written with the template file in its current state.

It's also worth noting that server.erb and client.erb should probably implement similar variable checking and standardising on one or other solution.


luxflux:
Thanks for your fix.
eah, you are right, the variable checking should be the same. I just didn't have the time refactor the module to have to handle this at one place :disappointed: 

Could you merge master (which fixes the travis build) and add a test for this?

typhonius:
@luxflux thanks for your feedback. I've rebased on master and added a test.

luxflux:
After merging your other feature this one has merge conflicts... Could you rebase once more? Thanks!

luxflux:
I am going to merge this pull request with some other changes via #129.

PR #118, opened 2015-01-08
Add TLS and x509-name verification support

maetthu-indermuehle:
- Server config
  - Added tls-server and tls-auth parameters to server config template
  - Generate TLS key on server while setting up the CA
- Client config
  - Link tls key into client config bundle
  - Added x509-name parameter to client config template
- Added test for tls and x509-name support in client and server

### DIFF
--- a/manifests/ca.pp
+++ b/manifests/ca.pp
@@ -103,7 +103,7 @@ define openvpn::ca(
   $key_cn = '',
   $key_name = '',
   $key_ou = '',
-  $tls_auth = 'false',
+  $tls_auth = false,
 ) {
 
   include openvpn
@@ -180,15 +180,6 @@ define openvpn::ca(
     require  => Exec["initca ${name}"],
   }
 
-  if $tls_auth {
-    exec { "generate tls key for ${name}":
-      command  => "openvpn --genkey --secret keys/ta.key",
-      cwd      => "/etc/openvpn/${name}/easy-rsa",
-      creates  => "/etc/openvpn/${name}/easy-rsa/keys/ta.key",
-      provider => 'shell',
-      require => Exec["copy easy-rsa to openvpn config folder ${name}"],
-    }
-  }
   file { "/etc/openvpn/${name}/keys":
     ensure  => link,
     target  => "/etc/openvpn/${name}/easy-rsa/keys",
@@ -207,6 +198,15 @@ define openvpn::ca(
     creates  => "/etc/openvpn/${name}/crl.pem",
     provider => 'shell',
     require  => Exec["generate server cert ${name}"],
+  }
+  if $tls_auth {
+    exec { "generate tls key for ${name}":
+      command  => "openvpn --genkey --secret keys/ta.key",
+      cwd      => "/etc/openvpn/${name}/easy-rsa",
+      creates  => "/etc/openvpn/${name}/easy-rsa/keys/ta.key",
+      provider => 'shell',
+      require  => Exec["generate server cert ${name}"],
+    }
   }
 
   file { "/etc/openvpn/${name}/easy-rsa/keys/crl.pem":
--- a/manifests/ca.pp
+++ b/manifests/ca.pp
@@ -54,6 +54,10 @@
 #     and KEY_CN in vars
 #   Default: None
 #
+# [*tls_auth*]
+#   Boolean. Determins if a tls key is generated
+#   Default: False
+#
 # === Examples
 #
 #   openvpn::ca {
@@ -99,6 +103,7 @@ define openvpn::ca(
   $key_cn = '',
   $key_name = '',
   $key_ou = '',
+  $tls_auth = 'false',
 ) {
 
   include openvpn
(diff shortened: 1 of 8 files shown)


manifests/ca.pp
```diff
@@ -99,6 +103,7 @@
   $key_cn = '',
   $key_name = '',
   $key_ou = '',
+  $tls_auth = 'false',
```
luxflux: Please use a boolean instead of a string here.

manifests/ca.pp
```diff
@@ -175,6 +180,15 @@
     require  => Exec["initca ${name}"],
   }
 
+  if $tls_auth {
+    exec { "generate tls key for ${name}":
+      command  => "openvpn --genkey --secret keys/ta.key",
+      cwd      => "/etc/openvpn/${name}/easy-rsa",
+      creates  => "/etc/openvpn/${name}/easy-rsa/keys/ta.key",
+      provider => 'shell',
+      require => Exec["copy easy-rsa to openvpn config folder ${name}"],
```
luxflux: Shouldn't this be run after `Exec["initca ${name}"]` or even after `Exec["generate server cert ${name}"]`?

manifests/client.pp
```diff
@@ -197,6 +210,15 @@
     target  => "/etc/openvpn/${server}/easy-rsa/keys/ca.crt",
     require => Exec["generate certificate for ${name} in context of ${server}"],
   }
+  if $tls_auth {
+    file { "/etc/openvpn/${server}/download-configs/${name}/keys/${name}/ta.key":
+      ensure  => link,
+      target  => "/etc/openvpn/${server}/easy-rsa/keys/ta.key",
+      require => Exec["generate certificate for ${name} in context of ${server}"],
+      before  => Exec["tar the thing ${server} with ${name}"],
```
luxflux: Please specify the dependency on the other side (`Exec["tar the thing ${server} with ${name}"]`) with `require`.

spec/defines/openvpn_client_spec.rb
```diff
@@ -78,7 +78,9 @@
       'auth_retry'            => 'interact',
       'verb'                  => '1',
       'setenv'                => {'CLIENT_CERT' => '0'},
-      'setenv_safe'           => {'FORWARD_COMPATIBLE' => '1'}
+      'setenv_safe'           => {'FORWARD_COMPATIBLE' => '1'},
+      'tls_auth'              => true,
+      'x509_name'             => 'test_server'
```
luxflux: Please add a trailing comma so we have less changes next time.

spec/defines/openvpn_client_spec.rb
```diff
@@ -103,6 +105,8 @@
     it { should contain_file('/etc/openvpn/test_server/download-configs/test_client/test_client.conf').with_content(/^setenv\s+CLIENT_CERT\s+0$/)}
     it { should contain_file('/etc/openvpn/test_server/download-configs/test_client/test_client.conf').with_content(/^setenv_safe\s+FORWARD_COMPATIBLE\s+1$/)}
     it { should contain_file('/etc/openvpn/test_server/download-configs/test_client/test_client.conf').with_content(/^cipher\s+BF-CBC$/)}
+    it { should contain_file('/etc/openvpn/test_server/download-configs/test_client/test_client.conf').with_content(/^tls-client$/)}
+    it { should contain_file('/etc/openvpn/test_server/download-configs/test_client/test_client.conf').with_content(/^verify-x509-name\s+"test_server"\s+name$/)}
```
luxflux: Could you add some reverse tests? So that it does not appear in the config if the parameter has not been specified?

spec/defines/openvpn_server_spec.rb
```diff
@@ -207,7 +210,8 @@
               :key_expire   => 365,
               :key_cn       => 'yolo',
               :key_name     => 'burp',
-              :key_ou       => 'NSA')
+              :key_ou       => 'NSA',
+              :tls_auth     => true)
```
luxflux: Please add a trailing comma so we have less changes next time.

spec/defines/openvpn_server_spec.rb
```diff
@@ -189,6 +190,8 @@
     it { should contain_file('/etc/openvpn/test_server.conf').with_content(%r{^down "/tmp/down"$}) }
     it { should contain_file('/etc/openvpn/test_server.conf').with_content(%r{^script-security 2$}) }
     it { should contain_file('/etc/openvpn/test_server.conf').with_content(%r{^duplicate-cn$}) }
+    it { should contain_file('/etc/openvpn/test_server.conf').with_content(%r{^tls-server$}) }
+    it { should contain_file('/etc/openvpn/test_server.conf').with_content(%r{^tls-auth\s+/etc/openvpn/test_server/keys/ta.key\s+0$}) }
```
luxflux: Could you add some reverse tests? So that it does not appear in the config if the parameter has not been specified?

luxflux:
Thanks a lot for this contribution! I added some comments directly in the code.

manifests/client.pp
maetthu-indermuehle: If I do this I have to copy the whole code since this file is only required as long as tls_auth is true, otherwise it is not required. Or do you have any other suggestion how to solve that nicely?

maetthu-indermuehle:
Thanks for the Input. I changed the points you mentioned. Except for the ordering. I don't have any good solution there without coping unnecessary code. Do you have any good ideas?

tobru:
:top:

manifests/client.pp
luxflux: You are right... My bad.

luxflux:
You are right.... Could you merge master once more as there are merge conflicts? Afterwards we are ready to merge.

maetthu-indermuehle:
I did a merge to the master.

luxflux:
Thanks, could you fix the tests?

maetthu-indermuehle:
I fixed the test. One parameter was set twice.

luxflux:
Thanks!

tobru:
Thanks for merging! =)

PR #122, opened 2015-01-15
added systemd style service name support for RHEL >= 7

tsujamin:
systemd based distributions use the "openvpn@instance" service name format. A switch has been added which detects these systems (currently just RHEL >=7) and adjusts the service name accordingly.

I'm not sure if there is a better place in the module to implement this change as the $name variable is required.

Tested on CentOS7

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -37,7 +37,6 @@ class openvpn {
   class {'openvpn::params': } ->
   class {'openvpn::install': } ->
   class {'openvpn::config': } ~>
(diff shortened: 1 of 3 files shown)


elisiano:
service_name is not a parameter of openvpn::service (hence the failure in travis)

luxflux:
Thanks for your pull request and for the work you've done here!
As #127 is a more complete solution and passes the tests, I am closing this one.

tsujamin:
awesome :)

PR #124, opened 2015-01-26
Added parameter (startall = true | false ) to allow start all configs

marianschmotzer:
... in openvpn directory, not only puppet managed.Works only on debian

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -138,6 +138,10 @@
 #   Boolean, Enable/Disable.
 #   Default: false
 #
+# [*startall*]
+#   Boolean, Define if to start also non pupet managed vpn confings on Debian                                                                                                                                                      
+#   Default: false
+#
 # [*management*]
 #   Boolean.  Enable management interface
 #   Default: false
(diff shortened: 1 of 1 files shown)


luxflux:
Thanks for your contribution. Unfortunately this won't work out as you implemented it. This way, it will create a `default` file with the following content if you specify multiple services:

```bash
AUTOSTART="all"
AUTOSTART="all"
AUTOSTART="all"
AUTOSTART="$AUTOSTART server123"
```

The last one would be a service which doesn't specify `start_all => true`...

This should be done on a central configuration class. Maybe we should add some parameters to `openvpn` directly in `init.pp`? What do you think?

marianschmotzer:
Hi
Maybe i can create facter to get line AUTOSTART=\"all\" from config. If it
is there we don't need to add any lines, because all configuration files
will be executed anyway.
And switch startall can stay in place (or if you would like to change name,
lease write me new one).
Or maybe another option will be to add parameter 'managestartconfig =>
true|false' with obvious function - puppet will or will not manage config.
Please tell me what do you think is better - my personal opinion is that
facter option is better.

Tnx

2015-02-02 18:53 GMT+01:00 Raffael Schmid <notifications@github.com>:

luxflux:
Thanks for your reply, @marianschmotzer. I have to dig into this first to find a good solution, stay tuned.

luxflux:
I went for the variable on `openvpn` and made a new pull request #144. Thanks for your work!

PR #127, opened 2015-01-29
support systemd-based distros

jlambert121:
The sysV init script loads all conf files from /etc/openvpn, the systemd service unit file takes a parameter for individual config files.  As a result, non-systemd systems need to support the "old" global service, systemd systems needs to have a service per server defined.

This PR supersedes #125 and can replace #122.  #98 is another possible implementation.


luxflux:
Thanks for your work! :beers:

As there are merge conflicts, I am going to merge this via #132.

PR #128, opened 2015-01-30
ensure unset client cipher is unset in config

jlambert121:


luxflux:
Thanks for your work! I am going to merge this pull request with some other changes via #129.

PR #133, opened 2015-02-04
normalize config, cleanup ensure_resource

jlambert121:
This PR has a lot of changes in it, almost all of them are minor spacing things to try to normalize the spacing between all resources.

The one larger change is to remove the ensure_resource on /etc/openvpn/${name} introduced with the shared_ca merge.  That seemed to be a large hack for a complicated dependency, but looks like it would be something to easily break later.  Needing to define a resource multiple times is usually an indication that something needs to be reworked.

I'm happy to break out any of these into different PRs (or undo any of the changes in this) if desired.

### DIFF
--- a/manifests/ca.pp
+++ b/manifests/ca.pp
@@ -114,25 +114,19 @@ define openvpn::ca(
   }
 
   File {
-    group   => $group_to_set,
+    group => $group_to_set,
   }
-
-  # directory shared with openvpn::server
-  ensure_resource(file, "/etc/openvpn/${name}", {
-    ensure  => directory,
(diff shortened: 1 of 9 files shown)


luxflux:
Thanks! Such a cleanup is always something I'd love to merge!

When looking over it, I saw some stuff I am not yet sure to merge as they might brake some things. For an easier QA, could you break out the spacing changes?

jlambert121:
Let me make some time for this, I figured that was going to be the request when I got done with this :)

jlambert121:
#135 and #136 breaks this out into 2 PRs

PR #136, opened 2015-02-12
refactor to remove ensure_resource

jlambert121:
The ensure_resource shared between server and ca seems like an anti-pattern.  I've refactored that out.

### DIFF
--- a/manifests/ca.pp
+++ b/manifests/ca.pp
@@ -117,22 +117,20 @@ define openvpn::ca(
     group => $group_to_set,
   }
 
-  # directory shared with openvpn::server
-  ensure_resource(file, "/etc/openvpn/${name}", {
-    ensure => directory,
-    mode   => '0750',
-  })
-
   exec { "copy easy-rsa to openvpn config folder ${name}":
     command => "/bin/cp -r ${openvpn::params::easyrsa_source} /etc/openvpn/${name}/easy-rsa",
     creates => "/etc/openvpn/${name}/easy-rsa",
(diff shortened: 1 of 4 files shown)


manifests/ca.pp
```diff
@@ -117,22 +117,16 @@
...
     command => "/bin/cp -r ${openvpn::params::easyrsa_source} /etc/openvpn/${name}/easy-rsa",
     creates => "/etc/openvpn/${name}/easy-rsa",
-    notify  => Exec["fix_easyrsa_file_permissions_${name}"],
     require => File["/etc/openvpn/${name}"],
   }
 
-  exec { "fix_easyrsa_file_permissions_${name}":
-    refreshonly => true,
-    command     => "/bin/chmod 750 /etc/openvpn/${name}/easy-rsa/*",
+  file { "/etc/openvpn/${name}/easy-rsa":
```
luxflux: This is about changing the permissions of the scripts directly inside `.../easy-rsa/`, see the asterisk in the command. With this new implementation, it would not change the permissions of the script but of the folder only.

Maybe a better solution would be to change the permissions of the used scripts only: `vars`, `clean-all`, `,build-dh` and `pkitool`. What do you think?

luxflux:
Thanks for fixing this! :+1: 

I made a comment directly in the code, could you check it? Also, could you merge `master` into your branch as there are some merge conflicts...

jlambert121:
Nice catch - I meant to add recurse => true.  I prefer the explicit definition though so went that route.

luxflux:
Thanks for fixing!

PR #138, opened 2015-03-04
feat (server): add fragment parameter support

brunoleon:
Conflicts:
	spec/defines/openvpn_server_spec.rb
	templates/server.erb


luxflux:
Thanks for your work, I am going to merge this via #141.

PR #139, opened 2015-03-04
Fix path to status_log to make it compatibler with SElinux RHEL7 policy

brunoleon:
Status log should be in /var/log for SELinux to accept it in enforcing mode.
This patch also moves the status log for debian like system.
It creates the folder as necessary

Conflicts:
	manifests/install.pp
	manifests/server.pp
	templates/server.erb


luxflux:
Thanks for your work, I merged it via #140.

PR #145, opened 2015-03-17
pam plugin path customization

catalinpan:
Hi

For x86_64 CentOS systems the openvpn pam plugin path is /usr/lib64/openvpn/plugin/lib/openvpn-auth-pam.so by default.

I have added a new variable which allows to have a custom path because the default one was for another system, I believe. Maybe this will help somebody in the future.

I have kept the old path as default /usr/lib/openvpn/openvpn-auth-pam.so

Thanks

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -138,6 +138,9 @@
 #   Boolean, Enable/Disable.
 #   Default: false
 #
+# [*pam_path*]
+#   Default: /usr/lib/openvpn/openvpn-auth-pam.so
+#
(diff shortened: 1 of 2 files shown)


luxflux:
Thanks, @catalinpan. I would to handle this in `params.pp`.
Do you mind adding a variable there and removing the parameter from `server.pp`? You can follow the pattern of `$easyrsa_source` in `params.pp`.

luxflux:
I had some time and did it myself. Will merge this via #148.

PR #146, opened 2015-03-20
Adding tls-client support and check logic

justicel:
Lots of lint cleanup, fix tcp server/client issues and added the ability to specify a custom hash of options for the server.


luxflux:
Thanks for your work, @justicel :beers:
I added some more tests and will merge it via #147.

PR #150, opened 2015-03-23
Update server.pp

justicel:
Apparently some people are having issues with the comment being in-line


luxflux:
Thanks! I don't think this will fix his bug, but it looks better this way anyway :smiley:

PR #155, opened 2015-04-24
Support for alternate OpenVPN config dir location

jlk:
In certain use cases, /etc/openvpn might be on a ephermal file system.
If VPN server cloud instances are being destroyed/re-created, having
the OpenVPN config directory on non-ephermal storage (eg an EBS drive
mounted at /home/openvpn) allows keys and certificates to survive
instance rebuilds.


luxflux:
Please merge `master`, it contains a fix for the failing build. Please also add some tests for your changes.

luxflux:
Closed due to inactivity.

PR #156, opened 2015-05-05
feature (server): add nobind parameter support.

andrekeller:
When the server is running in client mode, you have to specify
a fixed port for every vpn server. Alternativly when you set nobind
openvpn will just pick a random unused port.

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -314,6 +314,10 @@
 #   with client configuration
 #   Default: true
 #
+# [*nobind*]
+#   Boolean. Whether or not to bind to a specific port number.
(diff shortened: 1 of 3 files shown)


andrekeller:
Hi,

I'm not quite sure about the tests. Does this make sense for you? Currently when testing server mode, it was not tested that the port number is actually in the file. So maybe it would make more sense to define two separate test cases, one with nobind false that has port xxx in the file and one like it is currently.

g
André

luxflux:
Thank you for your work!
Regarding the test case, I will add one. Thats fine.

PR #157, opened 2015-05-05
feature: Add possibilty to not manage the openvpn service with puppet.

andrekeller:
When you have redundant pairs of firewalls you sometimes want to use
other means to start/stop the service (i.e. keepalived scripts).


andrekeller:
Ups this has some dependencies that I was not aware of... Will reopen once I have a better solution.

PR #158, opened 2015-05-05
feature: Add possibilty to not manage the openvpn service with puppet.

andrekeller:
When you have redundant pairs of firewalls you sometimes want to use
other means to start/stop the service (i.e. keepalived scripts).

This is primarly a request for comments. If somebody has a better approach, 
I'd be happy to dig into it.

### DIFF
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -30,10 +30,12 @@
 # limitations under the License.
 #
 class openvpn::service {
-  service { 'openvpn':
-    ensure     => running,
-    enable     => true,
-    hasrestart => true,
-    hasstatus  => true,
+  if $::openvpn::manage_service {
+    service { 'openvpn':
+      ensure     => running,
+      enable     => true,
+      hasrestart => true,
+      hasstatus  => true,
+    }
   }
 }
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -8,6 +8,9 @@
 # [*autostart_all*]
 #   Boolean. Wether the openvpn instances should be started automatically on boot.
 #   Default: true
+# [*manage_service*]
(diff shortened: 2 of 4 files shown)


manifests/service.pp
```diff
@@ -29,11 +34,15 @@
...
 #
-class openvpn::service {
-  service { 'openvpn':
-    ensure     => running,
-    enable     => true,
-    hasrestart => true,
-    hasstatus  => true,
+class openvpn::service ($manage_service = true) {
+
+  if $manage_service {
```
luxflux: Instead of adding a parameter, please reference `$::openvpn::manage_service` directly (as you do for the systemd part.

luxflux:
Well, the approach seems okay for me. I only had one comment, please check.

andrekeller:
I added it as parameter so I do not have to worry about the existing test case which invokes the service class directly. I'll give it another go.

luxflux:
You can handle this with a precondition:
```ruby
let(:pre_condition) { 'class { "openvpn": manage_service => true }' }
```

andrekeller:
Hmm does not seem to work that way... Any idea?

luxflux:
Expand the facts to be like this:

```ruby
let(:facts) do
  {
    :osfamily => 'Debian',
    :operatingsystem => 'Debian',
    :concat_basedir => '/var/lib/puppet/concat'
  }
end
```

andrekeller:
Now it looks better, thank you. 

Sorry this rspec stuff is pretty new to me, I do not yet fully understand the concepts and structure of these tests.

luxflux:
Thanks for your work! Looks good! I will add a reverse test to complete it.

PR #160, opened 2015-05-07
Add autostart_all check in default template

### LINKED ISSUE #159: `autostart_all` has no effect
Setting it to `true` does not change the `AUTOSTART` variable in `/etc/default/openvpn`.

jsok:
Fixes #159


luxflux:
Thanks for fixing the bug. I added some more changes and will merge it via #163 as soon as the tests are passing.

PR #164, opened 2015-05-12
feature: Add means to define openvpn server, clients etc. using hiera.

andrekeller:
Tested on Ubuntu 12.04 / 14.04

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,6 +16,10 @@ end
 
 if puppetversion = ENV['PUPPET_GEM_VERSION']
   gem 'puppet', puppetversion, :require => false
+  if puppetversion == "~> 2.7.0"
+    gem 'hiera-puppet', :require => false
+    gem 'hiera', :require => false
+  end
 else
   gem 'puppet', :require => false
 end
--- a/README.md
+++ b/README.md
@@ -76,6 +76,40 @@ Puppet module to manage OpenVPN servers
   }
 ```
 
+## Example with hiera
+
+```yaml
+---
+classes:
+  - openvpn
+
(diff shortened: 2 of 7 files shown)


andrekeller:
hmm this breaks puppet 2.7. (automagic parameter lookups / hiera_hash not supported).

andrekeller:
adding hiera-puppet module as dependency for puppet < 3.0.0 would probably solve this...

andrekeller:
this fixes the build error, however this is not tested on a real puppet 2.7 environment. Additionally maybe a notice should be placed in the changelog for puppet 2.7 users to install hiera-puppet gem / package on their puppet master.

luxflux:
I am not sure about this one. Do you know how widespread Puppet 2.7 is? We are using 3.x... Maybe we could drop Puppet 2.7 support altogether?

Regarding the feature itself, I have to check it a bit deeper. Never used hiera this way before...

andrekeller:
I have no idea how widespread Puppet 2.7 is, but there are definitely people out there still using it. 

Debian Wheezy (now oldstable, still supported), Ubuntu 12.04 LTS (still supported to 2017) ship with Puppet 2.7. I have no idea how many people use the shipped version though.

andrekeller:
Otherwise It would be a good excuse to bump the module's major version to 3 :-)

luxflux:
Yeah, this is my thought... I think I will release another 2.x version and merge this PR afterwards to work in direction of 3.0.0.

luxflux:
Just release version 2.9.0 which will be the last version on 2.x. Merged your work now, thanks!

PR #165, opened 2015-05-13
fix params for old centos/redhat versions

kam1kaze:
Currently EPEL repository contains the same version of easy-rsa package for all CentOS/RedHat versions:
http://pkgs.org/search/easy-rsa

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -20,28 +20,15 @@ class openvpn::params {
       $group            = 'nobody'
       $link_openssl_cnf = true
       $pam_module_path  = '/usr/lib64/openvpn/plugin/lib/openvpn-auth-pam.so'
+      $additional_packages = ['easy-rsa']
+      $easyrsa_source      = '/usr/share/easy-rsa/2.0'
 
       # Redhat/Centos >= 7.0
       if(versioncmp($::operatingsystemrelease, '7.0') >= 0) {
-        $additional_packages = ['easy-rsa']
-        $easyrsa_source      = '/usr/share/easy-rsa/2.0'
-        $systemd             = true
-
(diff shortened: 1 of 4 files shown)


luxflux:
I am wondering, does this change frequently? Do they add new packages to EPEL during maintenance of a OS version? I don't know this from Debian/Ubuntu. There is a feature freeze and these versions/packages will be kept. Only security fixes will be backported. How does this RedHat/CentOS handle?

kam1kaze:
From EPEL description:
>EPEL (Extra Packages for Enterprise Linux) is a volunteer-based community effort from the Fedora project to create a repository of high-quality add-on packages that complement the Fedora-based Red Hat Enterprise Linux (RHEL) and its compatible spinoffs, such as CentOS and Scientific Linux.

You have to enable It manually and it can be compared with universe repo in Debian. The new packages can be added anytime to this repo. Also it uses feature freeze for exists version, but the rules are not so strict as in base repos.

Also this part of code was incorrect from the beginning because EPEL packages are not related to  minor versions of distro. 

      # Redhat/Centos >= 6.4
      } elsif(versioncmp($::operatingsystemrelease, '6.4') >= 0) {
        $additional_packages = ['easy-rsa']
        $easyrsa_source      = '/usr/share/easy-rsa/2.0'
        $systemd             = false

      # Redhat/Centos < 6.4 >= 6
      } elsif(versioncmp($::operatingsystemrelease, '6') >= 0) {
        $easyrsa_source = '/usr/share/openvpn/easy-rsa/2.0'
        $systemd        = false

I don't know when easy-rsa package was added for EPEL 5  (CentOS ~= 5.*). But currently it is there :)

PR #168, opened 2015-06-22
Ubuntu 15.04

mattrenner:
Some relatively minor adjustments to add compatibility for the recently released Ubuntu 15.04. 
Ubuntu now uses systemd, so have added the appropriate flag in `params.pp`. 

Unfortunately there is an issue in the latest version of puppet (3.7.2 - I think) on Ubuntu 15.04 where it thinks upstart is the default init provider - so have also amended the service resource attribute in `server.pp` accordingly.  While it isn't ideal to manually specify the service provider I don't think it's an issue as it's within a control block to limit the resource only to systemd enabled distributions. 

Please let me know if this needs adjustments, or could be improved. Happy to add tests as necessary but am relatively new to puppet testing so wasn't sure where best to stick them! 

Cheers

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -58,8 +58,15 @@ class openvpn::params {
           }
         }
         'Ubuntu': {
+          # Version > 15.04, vivid
+          if(versioncmp($::operatingsystemrelease, '15.04') >= 0){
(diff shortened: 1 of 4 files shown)


luxflux:
Thank you for your work! It looks good!

PR #169, opened 2015-07-14
Mr

mklette:
Hi,

may I kindly ask you for reviewing these changes? Please let me know of any doubts you may have about merging and I'll try to complete the changes.

Thanks
Mathias

### DIFF
--- a/manifests/client.pp
+++ b/manifests/client.pp
@@ -267,7 +267,7 @@ define openvpn::client(
   $name_escaped = regsubst(regsubst($name, '\.', '\\.', 'G'), '@', '\\@', 'G')
   exec { "generate ${name}.ovpn in ${server}":
     cwd         => "/etc/openvpn/${server}/download-configs/",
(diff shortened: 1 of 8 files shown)


manifests/client.pp
```diff
@@ -264,9 +302,10 @@
     notify      => Exec["generate ${name}.ovpn in ${server}"],
   }
 
+  $name_escaped = regsubst(regsubst($name, '\.', '\\.', 'G'), '@', '\\@', 'G')
```
luxflux: Do I understand this line correctly as that the following replacements will be done?

* `.` => `\.`
* `@` => `\@`

manifests/revoke.pp
```diff
@@ -58,7 +58,7 @@
   Openvpn::Revoke[$name]
 
   exec { "revoke certificate for ${name} in context of ${server}":
-    command  => ". ./vars && ./revoke-full ${name} ; test $? -eq 2 && touch revoked/${name}",
+    command  => ". ./vars && ./revoke-full ${name}; echo \"exit $?\" | grep -qE '(error 23|exit (0|2))' && touch revoked/${name}",
```
luxflux: What kind of problem does this solve? Doesn't your version of `revoke-full` return an exit code of 2?

luxflux:
Thanks, @mklette. I added some comments inline, looking forward to your replies.

manifests/client.pp
mklette: This is true.
It was required for us as we use full e-mail addresses in resource title. The former version works fine as long as simple user names are used (plain alphanumeric). In our case, Perl's regular expression handling would treat those characters special and in turns wont match properly.

manifests/revoke.pp
mklette: This part you might want to skip because the problem arose only when working with OpenSSL 1.0.1-4 - available in Ubuntu Precise: This version doesn't exit with '2' but shows 'error 23' exactly as newer versions would do.
This problem most likely affects only a very limited set of installations, hence you could surely omit this change.
PS. As I read this now, it's not even correct.

luxflux: What do you mean is not correct? The last part of your comment about the "limited set"?

earsdown:
The tls-cipher feature would be nice.... but I guess that's what $custom_options is for...

manifests/revoke.pp
mklette: Sorry for the confusion, it's working correctly for Ubuntu Precise and Ubuntu Trusty with this fix.

luxflux:
Thank you, @mklette!

PR #170, opened 2015-08-28
External CA support

earsdown:
Hey,
Thanks for the great module. I'm working on getting external CA support working (e.g. for integration with FreeIPA and certmonger) and was hoping to have the feature merged. There's quite a lot of changes so I'm splitting it into multiple commits.
Cheers

PR #171, opened 2015-08-28
Fix path to pam module plugin on redhat >= 7

earsdown:
Looks like the path should be /usr/lib64/openvpn/plugins/openvpn-plugin-auth-pam.so.


luxflux:
Thank you for your work!
Looks like this PR is based on #170, I will review it after we merged #170.

luxflux:
@earsdown After that we have merged #192, could you revise this PR and check if it is still relevant?

luxflux:
Closing because of inactivity.

PR #174, opened 2015-09-15
allow disabling TLS client-side authentication.

rvalle:
It should not be mandatory to use client side TLS authentication.
It is very useful to authenticate the LDAP Server (by providing the CA) and encrypt the traffic.

Authenticating the client side adds significant complexity and the client is already authenticating via ldap bind.
This PR allows to disable client side TLS authentication when TLS is in use.

By default it is enabled so that puppet code is backward compatible.

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -216,6 +216,10 @@
 #   String. LDAP TLS authentication: path to the CA certificates.
 #   Default: None
 #
+# [*ldap_tls_client_auth_enable*]
+#   Boolean. LDAP TLS authentication: Enable if TLS client authentication is required
+#   Default: true
+#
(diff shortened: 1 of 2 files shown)


luxflux:
Thanks, @rvalle! Could you write tests for this?

rvalle:
I would need some time to get familiar with your test suite. I am somehow new to both Ruby and Puppet. (although catching up!)
I will review which tests there are and see if I can add an additional check.

luxflux:
Cool, thanks a lot!

luxflux:
If you have any questions or need clarification, just ask me.

luxflux:
Closing because of inactivity.

PR #176, opened 2015-09-22
Use concat for ovpn generation.

jkroepke:
Hi,

I refactor the generation of the ovpn file. Its support inline tls-auth now (see https://github.com/luxflux/puppet-openvpn/issues/175).

Concat Module is already required in metdata.json

Tested on Ubuntu 14.04 w/ puppet 3.8.2 and OpenVPN 2.3.7 for Windows (VPN client).

Also, the concat files are located in /var/lib/puppet/concat, which only read read by puppet group and root.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ spec/fixtures
 .gitconfig
 .bundle
 Gemfile.lock
+.idea
--- a/manifests/client.pp
+++ b/manifests/client.pp
@@ -267,7 +267,7 @@ define openvpn::client(
       require => Exec["generate certificate for ${name} in context of ${server}"],
       before  => [
         Exec["tar the thing ${server} with ${name}"],
-        Exec["generate ${name}.ovpn in ${server}"],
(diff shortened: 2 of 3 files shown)


luxflux:
Thank you for your work! :beers:  Looks quite good! :star2: 
What do you think, could you create a new definition called `openvpn::ovpn` and move all the `ovpn` stuff in there? Afterwards we can call `openvpn::ovpn` from `openvpn::client`.

jkroepke:
In my opinion, If you want to splitt openvpn::client and openvpn::ovpn, you should do it with tar, too. Maybe openvpn::client::ovpn for more a better style.

luxflux:
Yep, this sound right. Do you mind doing this? Or maybe just the `ovpn` part?

jkroepke:
While i'm trying this, I got some dependency cycles.

For example, the tblk file are require the ovpn file must be run before the tar command.

I think one complex class is good enough, Refactoring can be later, not?

luxflux:
Okay, I am fine with this.

luxflux:
Thanks a lot for your work!

PR #177, opened 2015-10-12
Support pre-shared static keys

rfletcher:
Add support for the OpenVPN 'secret' config option


luxflux:
Thank you for this contribution!
Do you mind adding some specs for this?

luxflux:
I renamed the file and added some specs in #186.

PR #152, opened 2015-04-04
FreeBSD compatability [WIP]

dennisdegreef:
Can I get some feedback on the work i've been doing on freebsd compatability?

Tasks:
- [X] Configurable etc-directory for `/etc` and `/usr/local/etc`
- [X] Configurable root-group, linux has `root`, bsd has `wheel`
- [X] rc-scripts in `/etc/rc.conf.d/`
- [X] Create separate `/usr/local/etc/rc.d/openvpn-${name}` rc-scripts from `openvpn::server`
- [X] Change `/etc/rc.conf.d/openvpn` to `/etc/rc.conf.d/openvpn-${name}`
- [ ] Fix up tests again

PR #178, opened 2015-10-16
FreeBSD compatability

dennisdegreef:
This is a second shot at this. Sorry for the previous inactivity.
This is related to #152

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -45,13 +45,4 @@ class openvpn::config {
       order   => '01',
     }
   }
-
-  if $::operatingsystem == 'FreeBSD' {
-    file { '/etc/rc.conf.d/openvpn.conf':
-      owner   => root,
-      group   => $root_group,
-      mode    => '0644',
-      content => 'openvpn_enable="YES"'
-    }
-  }
 }
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -34,7 +34,7 @@ class openvpn::config {
   if $::osfamily == 'Debian' {
     concat { '/etc/default/openvpn':
       owner => root,
-      group => root,
+      group => $root_group,
       mode  => '0644',
       warn  => true,
     }
@@ -43,6 +43,15 @@ class openvpn::config {
       content => template('openvpn/etc-default-openvpn.erb'),
       target  => '/etc/default/openvpn',
       order   => '01',
+    }
+  }
+
+  if $::operatingsystem == 'FreeBSD' {
+    file { '/etc/rc.conf.d/openvpn.conf':
+      owner   => root,
+      group   => $root_group,
+      mode    => '0644',
+      content => 'openvpn_enable="YES"'
     }
   }
 }
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -34,7 +34,7 @@ class openvpn::config {
   if $::osfamily == 'Debian' {
     concat { '/etc/default/openvpn':
       owner => root,
-      group => $root_group,
+      group => $openvpn::params::root_group,
       mode  => '0644',
       warn  => true,
     }
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -42,7 +42,6 @@ class openvpn::params {
       $group             = 'nogroup'
       $link_openssl_cnf  = true
       $pam_module_path   = '/usr/lib/openvpn/openvpn-auth-pam.so'
-      $load_main_service = true
 
       case $::operatingsystem {
         'Debian': {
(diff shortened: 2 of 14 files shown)


dennisdegreef:
I'm now trying to write my first specfiles so tests against FreeBSD can be added :)

luxflux:
Cool! Looking forward to them! Just ask if you need any help!

dennisdegreef:
I've never written a specfile before, and am not that familiar with Ruby, but I think this should cover the basics?

dennisdegreef:
Rebased because of a merge conflict

manifests/config.pp
```diff
@@ -34,7 +34,7 @@
   if $::osfamily == 'Debian' {
     concat { '/etc/default/openvpn':
       owner => root,
-      group => root,
+      group => $root_group,
```
luxflux: Shouldn't this be `$openvpn::params::root_group`?

manifests/params.pp
```diff
@@ -34,9 +37,12 @@
...
     'Debian': { # Debian/Ubuntu
-      $group = 'nogroup'
-      $link_openssl_cnf = true
-      $pam_module_path = '/usr/lib/openvpn/openvpn-auth-pam.so'
+      $etc_directory     = '/etc'
+      $root_group        = 'root'
+      $group             = 'nogroup'
+      $link_openssl_cnf  = true
+      $pam_module_path   = '/usr/lib/openvpn/openvpn-auth-pam.so'
+      $load_main_service = true
```
luxflux: What is this variable used for? I cannot find any reference to it? If it's needed, we have to add it to `Archlinux` and `RedHat` as well, don't we?

manifests/server.pp
```diff
@@ -447,8 +447,12 @@
     if $::openvpn::params::systemd {
       $lnotify = Service["openvpn@${name}"]
     } else {
```
luxflux: Can't we use `elsif` here? I think it becomes more readable:

```puppet
if $::openvpn::params::systemd {
  $lnotify = Service["openvpn@${name}"]

} elsif $::openvpn::params::namespecific_rclink {
  $lnotify = Service["openvpn_${name}"]

} else {
  $lnotify = Service['openvpn']
  Openvpn::Server[$name] -> Service['openvpn']

}
```

manifests/server.pp
```diff
@@ -574,9 +580,28 @@
...
         enable   => true,
         provider => 'systemd',
-        require  => [ File["/etc/openvpn/${name}.conf"], Openvpn::Ca[$ca_name] ]
+        require  => [ File["${etc_directory}/openvpn/${name}.conf"], Openvpn::Ca[$ca_name] ]
       }
     }
   }
 
+  if $::openvpn::params::namespecific_rclink {
+    file { "/usr/local/etc/rc.d/openvpn_${name}":
```
luxflux: Please use `$etc_directory` here as well.

manifests/server.pp
```diff
@@ -574,9 +580,28 @@
         ensure   => running,
         enable   => true,
         provider => 'systemd',
-        require  => [ File["/etc/openvpn/${name}.conf"], Openvpn::Ca[$ca_name] ]
+        require  => [ File["${etc_directory}/openvpn/${name}.conf"], Openvpn::Ca[$ca_name] ]
       }
     }
   }
 
+  if $::openvpn::params::namespecific_rclink {
```
luxflux: This clause does not take in account if `$::openvpn::manage_service` is `false`. At least the `service` part should be excluded.

luxflux:
Thank you for your work! It is already in a good shape! It would be nice if you could address my comments, so I can merge it afterwards.

manifests/params.pp
dennisdegreef: Due to the way things were written previously, it came in hand for disabling the 'main' Service, it's an inverted boolean option for `namespecific_rclink`, which will have a service per server.
I removed it in `f225835`

manifests/server.pp
dennisdegreef: Fixed in `58f95da` while adding a santity check for using systemd and rclinks

manifests/config.pp
dennisdegreef: Fixed in `09cc8918`

manifests/server.pp
dennisdegreef: Fixed in https://github.com/dennisdegreef/puppet-openvpn/commit/ed100549bb4fac3a10fff22984f9396a0bce8682

dennisdegreef:
I've addressed the feedback and opened a fresh rebased pull request in #180

PR #181, opened 2015-11-25
Earsdown/external ca support

lendico-seong:
Hi, I really liked the changes by @earsdown. I applied the requested fixes, so we can get this merged.

See #170 .

PR #184, opened 2015-11-30
Add parameters to specify custom ca, key and cert files

maxenced:
This PR adds 3 parameters to server class, which enables to use custom ca/key/cert, for exemple to use the ones provided by puppet : https://community.openvpn.net/openvpn/wiki/PuppetCA


luxflux:
Thanks a lot for your work!

Unfortunately, this PR interferes with #181. As #181 adds the same functionality in a more complete manner, I'd like to go with the changes in #181.

Issue #187, opened 2016-01-13
Importing the generated .ovpn file into OpenVPN Connect mobile app fails


duckpuppy:
The generated client .ovpn files have the following options set:
```
ca keys/<client-name>/ca.crt
cert keys/<client-name>/<client-name>.crt
key keys/<client-name>/<client-name>.key
```
Because the directives are present in the file, the mobile client will not import the .ovpn file due to the files not being present.  Simply removing those three lines from the generated .ovpn file allows it to import successfully since those keys are also contained inline in the .ovpn file itself.  If having those directives is necessary for some other reason, perhaps an option could be introduced to control the generation of those directives.

luxflux:
Unfortunately I am not a user of the `.ovpn` files. Maybe @elisiano can give us some insights?

elisiano:
Currently I'm using an older version of the module (prior to the ovpn refactoring). I'll try to update and report back. 
Although if what @duckpuppy says is true, we have an issue. There should not be any external file reference (that's the whole point of having .ovpn files).

elisiano:
So yeah, this is the issue: https://github.com/luxflux/puppet-openvpn/blob/master/manifests/client.pp#L336

It includes the original config in the ovpn file which contains references to the external files.

elisiano:
@duckpuppy did you have a chance to test the PR I submitted?

duckpuppy:
I haven't seen that a PR was submitted for this ticket.  I should have some
time tomorrow to try it out.

luxflux:
@duckpuppy did it work out?

luxflux:
I assume this is fixed by #190.

PR #190, opened 2016-03-04
Change how .conf and .ovpn files are generated

elisiano:
Prior to this, .ovpn would include the .conf file, which means the external references to files were also included.
Now the main template has been split into 2 files: client.erb (included in both .conf and .ovpn) and client_external_auth.erb (included only in .conf)
This should fix issue #187


luxflux:
With the assumption that no feedback is good feedback, I assume this fixes #187.

Thanks, @elisiano!

PR #191, opened 2016-03-25
Add missing directory

sileht:
Hi,

This change adds a missing 'file' resource required by the "exec[copy easy-rsa to openvpn config folder]"

Chers,

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -572,6 +572,13 @@ define openvpn::server(
     mode   => '0750',
     notify => $lnotify,
   }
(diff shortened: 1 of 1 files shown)


luxflux:
Hm, exactly these lines are also in [server.pp](https://github.com/sileht/puppet-openvpn/blob/master/manifests/server.pp#L506-L510). How does your node config look like?

sileht:
I have changed the way I have fixed the bug. I think it's more clear now.

luxflux:
Yes, it is. Thank you!

PR #192, opened 2016-04-20
Fix external ca support

axier:
Hi, I have been trying to contact to @lendico-seong but I haven't had any response. I have fixed the testing issue that her pull request has. I am testing the module and the external certificates works fine. Unfortunately, It is required to import the ca, server and dh certs from outside of the module. I mean, It is needed to copy the certs to the server and them point the path where the certificates are located.


luxflux:
Thank you, @axier, @earsdown and @lendico-seong for your work!

@axier, as the last one who did something on this topic, would you mind adding a short example to `README.md` which explains how to use the external CA?

PR #201, opened 2016-06-05
Fixes #200

### LINKED ISSUE #200: Error (...), Openvpn::Server[...]: has no parameter named 'extca_enabled'
Hi,

if working with exported ressources @@openpvn::client{} the above Error is thrown at the client, exporting the ressource, since at that client there is no openvpn::server at all...

tk-s:
Fixes #200 if working with exported_ressources like i am


luxflux:
Thank you for reporting and fixing the issue!

PR #206, opened 2016-08-08
Remove upper limit of concat verion module

pmoranga:


luxflux:
Thank you, but #205 already fixes this.

PR #207, opened 2016-08-10
allow setting easyrsa_source as param

lobeck:
This module only supports default packages for ubuntu.
Between Ubuntu 12 and 13 the modules and easyrsa packages splitted and pathes changed

So we've added a parameter to override this path to get newer packages working.

### DIFF
--- a/manifests/ca.pp
+++ b/manifests/ca.pp
@@ -58,6 +58,9 @@
 #   Boolean. Determins if a tls key is generated
 #   Default: False
 #
+# [*easyrsa_source*]
+#   String. Override default easyrsa path
+#
 # === Examples
 #
 #   openvpn::ca {
--- a/manifests/ca.pp
+++ b/manifests/ca.pp
@@ -104,6 +104,7 @@ define openvpn::ca(
   $key_name     = '',
   $key_ou       = '',
   $tls_auth     = false,
+  $easyrsa_source = $::openvpn::params::easyrsa_source,
 ) {
 
   include openvpn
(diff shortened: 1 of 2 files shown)


manifests/ca.pp
```diff
@@ -104,6 +107,7 @@
   $key_name     = '',
   $key_ou       = '',
   $tls_auth     = false,
+  $easyrsa_source = $::openvpn::params::easyrsa_source,
```
luxflux: Instead of hard coding this, I please handle this in `params.pp`. Or what's the reason to handle this here?

lobeck: That way it will take the current values from `params.pp` but allow you to override it in the class call or over hiera. I don't see how this could be done in `params.pp` as there is a lot of conditional logic for different operating systems. Additionally this is a pretty common way of overriding values or did i misunderstand your request?

luxflux: We already handle [different versions of Ubuntu in `params.pp`](https://github.com/luxflux/puppet-openvpn/blob/master/manifests/params.pp#L68-L87) and it's especially always about this path (which seems to change for every new release...), so there's everything in place to add the path for v13.04, IMHO?
Or do I misunderstand the issue?

luxflux: It also confuses me a bit as this works for me on 16.04?

lobeck: We're still on 12.04. Now i see what you're requesting, i'll take a second look at it!

PR #209, opened 2016-08-18
fix broken namespecific rclink

alxwr:
(Puppet 3.8.6)

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -704,7 +704,7 @@ define openvpn::server(
   if $::openvpn::params::namespecific_rclink {
     file { "/usr/local/etc/rc.d/openvpn_${name}":
       ensure => link,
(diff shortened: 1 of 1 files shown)


alxwr:
The checks for 1.9.3 fail with `An error occurred while installing json_pure (2.0.2), and Bundler cannot continue.`
Succeedes for Ruby 2.0

.gitignore
```diff
@@ -4,4 +4,6 @@ spec/fixtures
 .gitconfig
 .bundle
 Gemfile.lock
-.idea
\ No newline at end of file
+.idea
+*.swp
```
luxflux: These `gitignore` entries are a good fit for your global `~/.gitignore`, but please don't add them here.

alxwr: OK. I removed them.

luxflux:
Thank you!

PR #215, opened 2016-11-02
Add rhel6 support for ldap auth plugin

miguelwhite:
ref https://github.com/luxflux/puppet-openvpn/issues/172


luxflux:
Thank you!

PR #217, opened 2016-12-19
Correct path of openvpn-auth-pam.so on modern Debian distros.

oc243:
openvpn-auth-pam.so has been renamed to openvpn-plugin-auth-pam.so as of Saucy
[0] and Jessie [1]. This patch makes the module use the correct path for the
module after this change.

[0] http://packages.ubuntu.com/search?mode=exactfilename&suite=saucy&section=all&arch=any&keywords=openvpn-plugin-auth-pam.so&searchon=contents
[1] https://packages.debian.org/search?mode=path&suite=jessie&section=all&arch=any&searchon=contents&keywords=openvpn-plugin-auth-pam.so

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -42,7 +42,6 @@ class openvpn::params {
       $root_group          = 'root'
       $group               = 'nogroup'
       $link_openssl_cnf    = true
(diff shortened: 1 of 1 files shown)


oc243:
@luxflux don't suppose I could get a review on this if you get a free minute, pls :)

luxflux:
Merged, thank you!

PR #223, opened 2017-01-25
Turned up options for encryption

mcrmonkey:
These changes are to help this be "Secure by default"

+ Added: Default cipher for server and client. BF-CBC isn't that strong now and OVPN complains about this on startup
+ Added: 'timeout' for DH generation.Higher key sizes mean more time processing
+ Added: TLS ciphers added to help strengthen the server.

~ Changed: Key size now bumped up to 2048 which is now considered minimum key size requirement

TODO: TLS Ciphers need to be cleaned up and properly verified
TODO: It would be beneficial to install `haveged` to sure up the entropy pool especially during key generation


spec/defines/openvpn_server_spec.rb
```diff
@@ -95,7 +95,7 @@
     it { should_not contain_file('/etc/openvpn/test_server.conf').with_content(%r{^secret}) }
 
     it { should_not contain_file('/etc/openvpn/test_server.conf').with_content(/verb/) }
-    it { should_not contain_file('/etc/openvpn/test_server.conf').with_content(/cipher/) }
+    it { should contain_file('/etc/openvpn/test_server.conf').with_content(/cipher/) }
```
luxflux: Please test the value here.

spec/defines/openvpn_client_spec.rb
```diff
@@ -127,11 +127,6 @@
 
   end
 
-  context "omitting the cipher key" do
```
luxflux: Please test that the default is being added.

wyardley:
@mcrmonkey Are you still interested in advancing this, and if so, can you rebase (and optionally squash)?

mcrmonkey:
Ok I think this is a better place than it was before

Tests are now passing after fixing my bad merge attempt.

wyardley:
@mcrmonkey great! However, merging rather than rebasing introduces all the other commits -- at this point, it may make the most sense to make a patch against upstream/master, resync your master to upstream, make a feature branch, and apply the patch.... If you need help getting this squashed to one commit (or logical commits) against current master, feel free to reach out on the Voxpupuli Slack or IRC channels.

juniorsysadmin:
cc @roidelapluie

bastelfreak:
Looks good to me. @roidelapluie can you check it as well?

roidelapluie:
I will have a look

roidelapluie:
hard choice. definitively major bump for this. Do we really want that or rely on distro to make the right choice? 

also have you tested that those cifers are supported by the distributions we support?

wyardley:
Agree that there's some potential for drift - that is, if we create a more secure configuration, but then have to manage it, if the module doesn't get updated, we could actually create a _less_ secure configuration

roidelapluie:
Lots of this pull request is fine. Increasing sizes to 2048 is great. But hardcoding ciphers is something I would not recommend.

mcrmonkey:
Hopefully some answers below...

> rely on distro to make the right choice?

The openvpn config file is replaced by this module's output and IMO you're unlikely to see changes by the respective package maintainers to the openvpn code/resulting binary or unit file/init script that would override the default cipher selection. That would come from upstream.

Equally there are similar changes that could made to beef up the security of SSH setups.

> also have you tested that those cifers are supported by the distributions we support?

I've tested this on debian and centos machines and there were no issues with the ciphers used.

> Agree that there's some potential for drift - that is, if we create a more secure configuration, but then have to manage it, if the module doesn't get updated, we could actually create a less secure configuration

and

> hardcoding ciphers is something I would not recommend.

Both fair points. However times change and ciphers and cipher suites change with them. BF-CBC is a cipher thats thought of as weaker than the ones specified in this PR yet the same ciphers here will one day be obsolete so there will be follow up changes needed to keep them within the "secure" bracket.
Yet I feel doing nothing may be worse.

juniorsysadmin:
I made some line comments then stopped. There needs to be reference URLs in comments to back choices of defaults - https://www.keylength.com/en/4/ may be a helpful starting point. I am okay with this module not working by out of the box due to the choice of defaults provided (eg. TLS1.2-only ciphers), as long as this is clearly documented.

roidelapluie:
A module that does not work OOTB is never ok.

mcrmonkey:
Most of the arguments for and against the encryption options are covered in a readme file on another unrelated project whitch kind of does the same thing: https://github.com/Angristan/OpenVPN-install#encryption 
( saw this link on HN today which reminded me to provide more info for this PR )

https://github.com/Angristan/OpenVPN-install#data-channels-cipher covers information relating to the "cipher" option ( commonly know here as the 'BF-CBC' thing )

https://github.com/Angristan/OpenVPN-install#control-channels-cipher covers the additional cipher options and gives further detail on its supportability in other versions of openvpn

https://github.com/Angristan/OpenVPN-install#diffie-hellman-key touches on the size of the DH key. By default this module was generating a 1024 size key. NIST and the like do recommend larger DH key sizes as noted on the above link.

juniorsysadmin:
@mcrmonkey Linking to the mentioned links in the module comments and/or README would be great. Also, perhaps given the problem of trying to adopt a hardened default given older OpenVPN versions only support weak options it may be smarter to simply document that this module only supports installing OpenVPN version X or above.

wyardley:
@juniorsysadmin @roidelapluie @mcrmonkey 
are we on the same page about any additional changes needed? Do we want to merge this as-is, or ask for additional docs fixes?

juniorsysadmin:
If the links mentioned above are put in a README.md and/or code comments, I'm okay with merging.

wyardley:
@mcrmonkey Are you able to add the requested docs changes? Other than that, I think this one can be merged.

mcrmonkey:
I've updated the readme with the encryption choices and the reasoning behind them as well as adding a few references to the reasoning behind them.

bastelfreak:
Hi @mcrmonkey,
thanks for all the work in this PR. There was some discussion in this PR and you addressed all points, thanks for this. Contributions with docs + tests are rarely, so thanks for that!

Issue #232, opened 2017-05-26
error in File["${etc_directory}/openvpn/${name}"] dependencies


khaefeli:
`Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Invalid relationship: Exec[copy easy-rsa to openvpn config folder zh] { require => File[/etc/openvpn/zh] }, because File[/etc/openvpn/zh] doesn't seem to be in the catalog
`
the require comes from ca.pp line 124:
`    #require => File["${etc_directory}/openvpn/${name}"],`

edit: looks like the `openvpn::params` aren't loaded inside of the `openvpn::server` with puppet 4.9x and e.g. the etc_directory variable is empty.

if I inherit the params with the openvpn class, everything works like expected

Dan33l:
It should not be present with the current version of module.
You are welcome to reopen if the issue is yet present.

PR #233, opened 2017-05-26
fix wrong params for the server class

khaefeli:
fix for #232

### DIFF
--- a/lib/facter/client_crt.rb
+++ b/lib/facter/client_crt.rb
@@ -1,5 +1,6 @@
 require 'facter'
 
+if File.directory?('/etc/openvpn/')
 openvpn_clients = Dir['/etc/openvpn/*/easy-rsa/keys/*.crt'].map { |a| a.gsub(%r{\.crt$}, '').gsub(%r{^.*/}, '') }
 openvpn_ca = Dir.entries('/etc/openvpn').select {|entry| File.directory? File.join('/etc/openvpn',entry) and !(entry =='.' || entry == '..' || entry == 'keys') }
 
@@ -46,4 +47,5 @@ openvpn_clients.each do |openvpn_client|
       key
     end
   end
+end
 end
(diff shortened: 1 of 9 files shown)


lib/facter/client_crt.rb
```diff
@@ -0,0 +1,51 @@
...
+openvpn_ca = Dir.entries('/etc/openvpn').select {|entry| File.directory? File.join('/etc/openvpn',entry) and !(entry =='.' || entry == '..' || entry == 'keys') }
+
+
+Facter.add(:openvpn_clients) do
+  setcode do
+    openvpn_clients.join(',') if openvpn_clients
+  end
+end
+
+Facter.add(:openvpn_ca) do
```
ekohl: This and the next 3 facts should handle the case where ```openvpn_ca``` is empty I think.

lib/puppet/type/vpnserver.rb
```diff
@@ -0,0 +1,33 @@
+module Puppet
+Type.newtype(:vpnserver) do
+  @doc = "Install certificates for the openvpn server slave. 
+  Copied from the master over PuppetDB"
```
ekohl: Isn't this an provider detail rather than the type?

manifests/client.pp
```diff
@@ -218,7 +218,9 @@
   $pull                 = false,
   $server_extca_enabled = false
 ) {
-
+  err ($::openvpn_clients)
```
ekohl: This looks like test code.

manifests/slave.pp
```diff
@@ -0,0 +1,50 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+define openvpn::slave(
+) {
+  include openvpn
+  Class['openvpn::install'] ->
+  Openvpn::Slave[$name]
+
+  Vpnserver <<| |>>
```
ekohl: I can imagine that some users might have multiple sets of VPN servers so you might not want to collect everything.

lib/facter/client_crt.rb
```diff
@@ -0,0 +1,51 @@
+require 'facter'
+
+if File.directory?('/etc/openvpn/')
+openvpn_clients = Dir['/etc/openvpn/*/easy-rsa/keys/*.crt'].map { |a| a.gsub(%r{\.crt$}, '').gsub(%r{^.*/}, '') }
```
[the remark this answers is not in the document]
khaefeli: @ekohl thanks for reviewing. 
Sorry, I've created a PR from my master by mistake. Should be another branch for the params fix. (without my facter tests ;) )

PR #237, opened 2017-10-04
Relax on the concat version requirement

hp197:
Tested with puppet-concat 4.0.1 and works like a charm


metadata.json
```diff
@@ -18,7 +18,7 @@
     { "operatingsystem": "FreeBSD" }
   ],
   "dependencies": [
-    {"name":"puppetlabs/concat","version_requirement":">= 1.0.1 <3.0.0"},
+    {"name":"puppetlabs/concat","version_requirement":">= 1.0.1 <=4.0.1"},
```
dhollinger: Is there any reason to not push the concat version req range out to `< 5.0.0`?

wyardley:
This may need to be updated after #239 is merged, since I reformatted the metadata. I think the next release after the first Vox release will have dependency version bumps across the board, and will add an explicit Puppet >= 4.7.1 requirement in the metadata.

metadata.json
hp197: Yes and no, I can only say I have tested with 4.0.1
That being said, minors shouldn't introduce breaking changes....

wyardley:
@hp197 can you rebase or re-do? That said, I could also just bump the metadata for you when I do some other adjustments to the versions / dependencies.

hp197:
i'll redo

wyardley:
@hp197 awesome. I think it's good to just bump both concat and stdlib to require current versions.
Something like this is probably reasonable (but with lower end of Puppet version 4.7.1 instead of 4.7.0).
https://github.com/voxpupuli/puppet-nginx/blob/master/metadata.json#L10-L25

hp197:
rebased back to master without squash (never squashed before, was unsure about it).
Ran the tests and all where ok.

I added support for Debian 9 (stretch), as we run on it and it works.

wyardley:
@hp197 I think you have some other commits in here, maybe you merged instead of rebasing? Also, there's a conflict again.

wyardley:
@hp197 Sorry, but I bumped the version along with some other versions in #242, so assuming that gets merged, this shouldn't be necessary.

hp197:
Yeah, like I said im blunt with rebasing...
Let me close this PR and open a new one based on the current master if I still need adjustments

PR #238, opened 2017-10-05
modulesync v1.2.0

luxflux:
Add all the voxpupuli modulesync stuff


spec/default_module_facts.yml
```diff
@@ -0,0 +1,6 @@
+id: root
+path: 
+ipaddress_eth0: 10.0.0.1
+osfamily: Debian
```
wyardley: I think these should already be in (or should be added in?) the os specific stuff in the spec tests instead?

And we should switch to rspec-puppet-facts (but as a separate thing)

PR #242, opened 2017-10-07
Breaking: Update puppet, stdlib, and concat requirements in prep for release

wyardley:
Open to suggestions on whether to do another release immediately or to merge some stuff first.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,11 +1,7 @@
 fixtures:
   repositories:
-    stdlib:
-      repo: git://github.com/puppetlabs/puppetlabs-stdlib
-      ref: '4.6.0'
-    concat:
-      repo: git://github.com/puppetlabs/puppetlabs-concat
-      ref: '1.2.1'
+    stdlib: "git://github.com/puppetlabs/puppetlabs-stdlib"
+    concat: "git://github.com/puppetlabs/puppetlabs-concat"
   symlinks:
     openvpn: "#{source_dir}"
 
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Puppet module to manage OpenVPN servers and clients.
 * Amazon
 
 ## Dependencies
-  - [puppetlabs-concat 4.0.0+](https://github.com/puppetlabs/puppetlabs-concat)
+  - [puppetlabs-concat 3.0.0+](https://github.com/puppetlabs/puppetlabs-concat)
   - [puppetlabs-stdlib 4.13.1+](https://github.com/puppetlabs/puppetlabs-stdlib)
 
 ## Puppet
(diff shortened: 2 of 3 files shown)


wyardley:
We should probably switch to P4 data types + switch empty strings to undefs before the release.

bastelfreak:
can you check the mergeconflicts?

wyardley:
@bastelfreak: rebased. Also have a WIP with data types, but have a few questions to address.

.fixtures.yml
```diff
@@ -2,10 +2,10 @@ fixtures:
   repositories:
     stdlib:
       repo: git://github.com/puppetlabs/puppetlabs-stdlib
-      ref: '4.6.0'
+      ref: '4.20.0'
```
bastelfreak: As on the other PR: We should unpin those modules.

wyardley: updated

metadata.json
```diff
@@ -46,14 +46,20 @@
...
-      "name": "puppetlabs/concat",
-      "version_requirement": ">= 1.0.1 < 5.0.0"
+      "name": "puppetlabs/stdlib",
+      "version_requirement": ">= 4.13.1 < 5.0.0"
     },
     {
-      "name": "puppetlabs/stdlib",
-      "version_requirement": ">= 1.0.0 < 5.0.0"
+      "name": "puppetlabs/concat",
+      "version_requirement": ">= 4.0.0 < 5.0.0"
```
ekohl: And reason this wouldn't work with an older concat? In general it's nice to have the option to use an older version in case other modules aren't allowing ```4.0.0``` yet.

wyardley: @ekohl It probably still would -- however, once we stop testing against it, may be hard to assert that it still works with it. In general, we tend to drop support for older versions of stuff in modules, but I'm happy to set it to whatever makes sense.
nginx is using `">= 3.0.0 <5.0.0"`

wyardley: @bastelfreak do you want me to back it down to 3.0?

bastelfreak: 3.0 seems to be fine. 4.0 just added datatypes, but has the same functionality

README.md
```diff
@@ -29,12 +29,12 @@ Puppet module to manage OpenVPN servers and clients.
 * Amazon
 
 ## Dependencies
-  - [puppetlabs-concat 1.0.1+](https://github.com/puppetlabs/puppetlabs-concat)
-  - [puppetlabs-stdlib 1.0.0+](https://github.com/puppetlabs/puppetlabs-stdlib)
+  - [puppetlabs-concat 4.0.0+](https://github.com/puppetlabs/puppetlabs-concat)
```
bastelfreak: @wyardley can you add here 3.0.0 as well pls?

wyardley: good catch.

wyardley:
@bastelfreak is this looking good now?

PR #243, opened 2017-10-09
Upped version requirement of concat and added Debian 9 (stretch)

hp197:
Now with clean set of patches


bastelfreak:
Thanks for the PR @hp197!

PR #244, opened 2017-10-09
[WIP] Switch to Puppet data types

wyardley:
This adds Puppet 4 data types.


.fixtures.yml
```diff
@@ -2,10 +2,10 @@ fixtures:
   repositories:
     stdlib:
       repo: git://github.com/puppetlabs/puppetlabs-stdlib
-      ref: '4.6.0'
+      ref: '4.20.0'
```
bastelfreak: IMO we should unpin the modules here. In general we test against master to find breaking changes before they are released.

wyardley: @bastelfreak works for me, I'll adjust in #242

spec/defines/openvpn_server_spec.rb
```diff
@@ -166,7 +166,7 @@
         'key_cn'          => 'yolo',
         'key_name'        => 'burp',
         'key_ou'          => 'NSA',
-        'verb'            => 'mute',
```
wyardley: @luxflux do you know why verb is 'mute' here? It seems like `mute` is a separate param, and the docs suggest that `verb` should be an int as in the client?

luxflux: Yeah, this seems to be wrong. Let's use an integer...
From https://openvpn.net/index.php/open-source/documentation/manuals/65-openvpn-20x-manpage.html:
> --verb n
  Set output verbosity to n (default=1). Each level shows all info from the previous levels. Level 3 is recommended if you want a good summary of what's happening without being swamped by output.
  0 -- No output except fatal errors. 
  1 to 4 -- Normal usage range. 
  5 -- Output R and W characters to the console for each packet read and write, uppercase is used for TCP/UDP packets and lowercase is used for TUN/TAP packets. 
  6 to 11 -- Debug info range (see errlevel.h for additional information on debug levels).

I think we need we need to update this line as well: https://github.com/wyardley/puppet-openvpn/blob/7b4d56a6d048956dd8f706449f2158966745b68f/spec/defines/openvpn_server_spec.rb#L223

wyardley:
@luxflux edited that other line as well. The bulk of the tests currently fail with the same error, if you get a sec, could you see what I need to adjust?
```
Puppet::Error:
        Could not find resource 'Openvpn::Ca[winterthur]' for relationship on 'Openvpn::Client[winti-client1]'
```

I think the ordering on / from the base class is not needed, though maybe that's part of the problem?

luxflux:
@wyardley I'm not used to the Puppet data types. But what I can say is that a `openvpn::client` needs a `openvpn::ca` to work (directories, signing CA, ...) and that there is quite some magic involved here because a `openvpn::ca` may be used by multiple servers. To cover this, there is this [`pick` statement in `client.pp`](https://github.com/wyardley/puppet-openvpn/blob/431a88254fca88b634be91056f33a8dfb6b96764/manifests/client.pp#L232-L234).

Does this help?

wyardley:
@juniorsysadmin any thoughts on the test failures here?

juniorsysadmin:
@wyardley Hmm, I was going to suggest re-arrangement of `pre_condition` to be first in the defined types but that shouldn't have an effect. I take it you have tried sprinkling

```
Puppet::Util::Log.level = :debug
Puppet::Util::Log.newdestination(:console)
```
in the rspec, and turning off parallel rspec to see if that is helpful in finding the problem?

manifests/ca.pp
```diff
@@ -90,26 +90,26 @@
...
-  $common_name  = 'server',
-  $group        = false,
-  $ssl_key_size = 1024,
-  $ca_expire    = 3650,
-  $key_expire   = 3650,
-  $key_cn       = '',
-  $key_name     = '',
-  $key_ou       = '',
-  $tls_auth     = false,
+  String $country,
```
bastelfreak: datatypes \o/

manifests/server.pp
```diff
@@ -581,12 +581,24 @@
 
   if $extca_enabled {
     # VPN Server or Client with external CA
-    if $extca_ca_cert_file == undef { fail('extca_ca_cert_file has to be specified in extca mode') }
-    if $extca_ca_crl_file == undef and $crl_verify and !$remote { fail('extca_ca_crl_file has to be specified in extca mode if crl_verify is enabled') }
-    if $extca_server_cert_file == undef { fail('extca_server_cert_file has to be specified in extca mode') }
-    if $extca_server_key_file == undef { fail('extca_server_key_file has to be specified in extca mode') }
-    if $extca_dh_file == undef and !$remote and $tls_server { fail('cant enable tls_server: missing extca_dh_file') }
-    if $extca_tls_auth_key_file == undef and !$remote and $tls_auth { fail('cant enable tls_auth: missing extca_tls_auth_key_file') }
+    if $extca_ca_cert_file == undef {
```
bastelfreak: How do you like this?

```puppet
unless $extca_ca_cert_file {}
```

 This is maybe easier to read?

wyardley: I prefer that syntax myself, though I know others don't. I'm wondering if those sort of changes are best made after we at least get the tests passing w/ the updated code, and maybe in a separate PR?

wyardley: I was actually having the exact same thought, though, when I was rebasing it.

bastelfreak: yeah, we should not clutter up this PR.

wyardley:
I've rebased again, however, not sure how much time I'll be able to devote to this (it's not a module I use), and the tests still have the same failures. Is anyone interested in taking a deeper look into the test failures, or should I close this?

wyardley:
I'm going to close this; agree with some comments that this is too much all at once, and not that familiar with the module, so I'll let someone else pick up this torch if they're interested.

PR #251, opened 2017-11-20
replace validate_* with datatypes in init.pp

bastelfreak:

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -15,7 +15,7 @@
 #   Hash of defaults for clients passed to openvpn::client defined type.
 #   Default: {} (hiera_hash)
 # [*clients*]
-#   Hash of clients passed to openvpn::client defined type. 
+#   Hash of clients passed to openvpn::client defined type.
 #   Default: {} (hiera_hash)
 # [*client_specific_config_defaults*]
 #   Hash of defaults for client specific configurations passed to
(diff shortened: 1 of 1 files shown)


manifests/init.pp
```diff
@@ -69,27 +69,18 @@
...
-  $clients                         = hiera_hash('openvpn::clients', {}),
-  $client_specific_config_defaults = hiera_hash('openvpn::client_specific_config_defaults', {}),
-  $client_specific_configs         = hiera_hash('openvpn::client_specific_configs', {}),
-  $revoke_defaults                 = hiera_hash('openvpn::revoke_defaults', {}),
-  $revokes                         = hiera_hash('openvpn::revokes', {}),
-  $server_defaults                 = hiera_hash('openvpn::server_defaults', {}),
-  $servers                         = hiera_hash('openvpn::servers', {}),
+  $autostart_all                        = true,
+  $manage_service                       = true,
+  Hash $client_defaults                 = hiera_hash('openvpn::client_defaults', {}),
```
ekohl: What's the added benefit of ```hiera_hash``` here?

bastelfreak: I've no idea so I didn't want to touch it. Is this for strange hiera lookups? I guess this is obsolete, but should be addressed in a separate pull request.

ekohl: I think it's redundant but agree it's fine to do it in another PR (hence my approval rather than a comment/request).

PR #259, opened 2018-01-04
add datatypes to all params

to-kn:
Hi @bastelfreak, 
since you wished proper datatypes at my other PR, i just went through the code and added them for all params. Feel free to suggest any corrections, or to discard it.
Tests should be running fine


manifests/ca.pp
```diff
@@ -89,27 +89,27 @@
...
-  $group        = false,
-  $ssl_key_size = 2048,
-  $ca_expire    = 3650,
-  $key_expire   = 3650,
-  $key_cn       = '',
-  $key_name     = '',
-  $key_ou       = '',
-  $tls_auth     = false,
+define openvpn::ca (
+  String $country,
```
bastelfreak: Datatypes \o/

manifests/ca.pp
```diff
@@ -89,27 +89,27 @@
...
+  Integer $key_expire     = 3650,
+  String $key_cn          = '',
+  String $key_name        = '',
+  String $key_ou          = '',
+  Boolean $tls_auth       = false,
 ) {
 
   include openvpn
 
   $group_to_set = $group ? {
```
bastelfreak: Is this actually needed? Can't we just assign $openvpn::params::group in the params section above_ (line 99)

manifests/client.pp
```diff
@@ -236,7 +236,7 @@
   $etc_directory = $::openvpn::params::etc_directory
 
   if $expire {
-    if is_integer($expire){
+    if is_integer($expire) {
```
bastelfreak: In line 216 you marked it as boolean, so it can never be an interger.

bastelfreak:
Hi @to-kn, thanks for the PR. I added some inline comments.

manifests/ca.pp
to-kn: sadly yes (or at least i did not manage to do it) because if i do it like this:
"Optional[String] $group                   = $::openvpn::params::group,"
ill get an error like:
"Evaluation Error: Unknown variable: '::openvpn::params::group'"

to-kn: if you do have a better idea, i am glad to change it!

to-kn:
PS: travis failed only due to an travis-timeout-error...

manifests/ca.pp
bastelfreak: ahh, because this is a defined type. yeah let's stick with this logic.

bastelfreak:
Code looks good, I restarted the failed job.

Issue #264, opened 2018-02-18
Adding tls-version-min


peterbeck:
Hey there,

I am a longtime user of this module. But I've used quiet an old version so I did an upgrade, as so much was added / changed. Most of my servers are still running with Debian Jessie (8.x, soon to be upgraded).

It seems I can't connect with the current defaults - getting following message
```
SSL routines:SSL3_GET_CLIENT_HELLO:no shared cipher
```
when commenting tls-cipher in the server- and client config, it does work, but of course this is not what I intend to do.

But when adding tls-version-min on the server config, it seems to work flawless again
```
tls-version-min 1.2
```

Could we add these options for the server config ? Shall I create a pull request with that addition ?
As this is also recommended in the [hardening manual](https://community.openvpn.net/openvpn/wiki/Hardening), I think this would be reasonable.

Regards
Peter

bastelfreak:
Hi @peterbeck, thanks for this issue. Can you provide a PR for this?

peterbeck:
Hi @bastelfreak, of course, will do tomorrow. What do you think, defaults to an empty string / unset, so that existing deployments aren't impaired ?

peterbeck:
I've made a pull request - hope I've done it the right way ;-)

PR #265, opened 2018-02-20
added tls-version-min

peterbeck:
added tls-version-min as discussed in #264. this is my first PR - I hope I've done it correct...

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -257,6 +257,13 @@
 #   String, TLS Ciphers to use
 #   Default: TLS-DHE-RSA-WITH-AES-256-GCM-SHA384:TLS-DHE-RSA-WITH-AES-256-CBC-SHA256:TLS-DHE-RSA-WITH-AES-128-GCM-SHA256:TLS-DHE-RSA-WITH-AES-128-CBC-SHA256
 #
+# [*tls_version_min*]
+#   String, enforce a minimum TLS version
+#     As of OpenVPN 2.3.3, OpenVPN supports TLS version negotiation.
+#     Used to enforce a minimum TLS version. Be aware pre-2.3.3 clients 
+#     won't be able to connect if minimum version is set greater than 1.1
+#   Default: None / unset
+#
 # [*persist_key*]
 #   Boolean.  Try to retain access to resources that may be unavailable
 #     because of privilege downgrades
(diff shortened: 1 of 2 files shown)


manifests/server.pp
```diff
@@ -500,6 +507,7 @@
   String $verb                              = '',
   String $cipher                            = 'AES-256-CBC',
   String $tls_cipher                        = 'TLS-DHE-RSA-WITH-AES-256-GCM-SHA384:TLS-DHE-RSA-WITH-AES-256-CBC-SHA256:TLS-DHE-RSA-WITH-AES-128-GCM-SHA256:TLS-DHE-RSA-WITH-AES-128-CBC-SHA256',
+  String $tls_version_min                   = '',
```
bastelfreak: Datatypes \o/

bastelfreak: Can you change the logic so it defaults to undef instead of empty string? That would be the current best practice.

bastelfreak:
Hi @peterbeck, thanks for the PR. I'm wondering if we should provide a secure default setting for this option, or if it should default to nothing. I'm not very familiar with openvpn so I don't know which openvpn versions support which TLS versions and how easy it is to implement this. Do you have any ideas?

----

Can you please add a unit test that verifies that the config file contains the correct line?

peterbeck:
Hi @bastelfreak, thanks for your feedback. I think we should leave it to 'undef' by default not to break existing deployments, otherwise older (<2.3.4) clients [won't be able to connect ](https://community.openvpn.net/openvpn/ticket/401). But I'm also not that sure in this case.. I'll change the empty string and try to add a unit test (have to read first...)

bastelfreak:
Alright, lets stick with undef.

alexjfisher:
@peterbeck Have you had chance to look at making the change to use 'undef' and adding a test?

peterbeck:
I'm just asking myself if this pull request is necessary, as this parameter could also be set with custom_options.... 
```
custom_options => {
  "tls-version-min" => "1.2"
}
```
so far it seems I was the only one confronted with issues while not setting tls-min-version. 
what do you think ?

simondeziel:
@peterbeck, I see value in your PR and Optional[String] seems the way to go IMHO.

Dan33l:
Any news in the PR ?

Now a rebase is needed.

bastelfreak:
Hi people. This PR looks stalled, I'm going to close it. Please reopen and rebase it if you're still interested in it.

PR #270, opened 2018-03-14
Bug Fix: Ensure cipher and tls_cipher can be disabled entirely

jcarr-sailthru:
This change ensures that by setting cipher and tls_cipher to blank strings, they are consistently kept out of the configuration files - as opposed to invalid configuration being generated where the config keyword is generated with a blank argument.

Note that this does not change any defaults, since the Puppet module sets cipher and tls_cipher by default. Booleans can't be used, since we validate that these arguments are strings and boolean false is not a string :-)

### DIFF
--- a/templates/client.erb
+++ b/templates/client.erb
@@ -22,10 +22,10 @@ persist-key
 <% if @persist_tun -%>
 persist-tun
 <% end -%>
-<% if @cipher -%>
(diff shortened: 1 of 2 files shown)


bastelfreak:
Thanks for the PR @jcarr-sailthru !

PR #275, opened 2018-04-05
Allow management_port to be a string; require stdlib >= 4.25.0

### LINKED ISSUE #274: Configuring management unix socket is no longer possible

```
openvpn::server { 'name':
  management       => true,
  management_ip    => '/var/run/openvpn-name.sock',
  management_port  => 'unix',
}
```

## What are you seeing

```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Openvpn::Server[name]:
  parameter 'management_port' expects an Integer value, got String
```

Either it should still be possible set `management_port` to 'unix' or a new `management_socket` parameter which implies the unix port should be added.

marieof9:
Allow the `management_port` parameter to be a string too. This fixed #274


manifests/server.pp
```diff
@@ -478,7 +478,7 @@
   String $pam_module_arguments              = 'login',
   Boolean $management                       = false,
   String $management_ip                     = 'localhost',
-  Integer $management_port                  = 7505,
+  Variant[Integer,String] $management_port  = 7505,
```
bastelfreak: Is there any string valid, or only 'unix'? Otherwise we could do something like `Variant[Integer,Enum['unix']]`.

---

Could you also enhance this by removing the `Integer` type and replace it with `Stdlib::Port::Unprivileged` from https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/types/port/unprivileged.pp ? This requires a bump of the minimal stdlib version to 4.25.0 in the metadata.json:
https://github.com/voxpupuli/puppet-openvpn/blob/4814cd7529f707f675ef4adff4ff76431303a938/metadata.json#L56

bastelfreak:
thanks!

PR #276, opened 2018-05-07
Remove deprecated hiera_hash

jkroepke:
Hi,

since puppet has a [automatic class paramter lookup](https://puppet.com/docs/puppet/5.5/hiera_automatic.html), `hiera_hash` or even `lookup` isn't required.

This change is *maybe* not BC.


jkroepke:
Looks like tests are broken on this repository.

bastelfreak:
Thanks for the PR @jkroepke! We rebased it in https://github.com/voxpupuli/puppet-openvpn/pull/289

PR #278, opened 2018-05-11
Add Support for Easy RSA 3.x in EL7

NoodlesNZ:
Adding onto the initial work that @jameskirsop did to get this working with Easy RSA 3.x.

Added additional fixes including client config creation and server certificate/key paths.


templates/vars-30.erb
```diff
@@ -0,0 +1,80 @@
+# easy-rsa parameter settings
```
bastelfreak: as per our [review guidelines](https://voxpupuli.org/docs/#reviewing-a-module-pr), new templates should be epp.

manifests/ca.pp
```diff
@@ -124,20 +125,22 @@
...
 
-  file { [
-    "${etc_directory}/openvpn/${name}/easy-rsa/clean-all",
-    "${etc_directory}/openvpn/${name}/easy-rsa/build-dh",
-    "${etc_directory}/openvpn/${name}/easy-rsa/pkitool",
-  ]:
-    ensure  => file,
-    mode    => '0550',
-    require => Exec["copy easy-rsa to openvpn config folder ${name}"],
+  if $openvpn::params::easyrsa_ver == '2.0' {
```
bastelfreak: can we somehow determine the actual rsa version from the system?  One idea would be to parse `easyrsa --help` which prints `Easy-RSA 3` at least on my arch system

templates/vars-30.erb
NoodlesNZ: No problem, have converted this

manifests/ca.pp
NoodlesNZ: It might be possible to do this, but it would slow down puppet runs to run exec each time. Setting this via the params based on OS and version should be good enough, it's unlikely to change often.

manifests/ca.pp
```diff
@@ -124,20 +125,22 @@
     mode   => '0750'
   })
 
-  exec { "copy easy-rsa to openvpn config folder ${name}":
-    command => "/bin/cp -r ${openvpn::params::easyrsa_source} ${etc_directory}/openvpn/${name}/easy-rsa",
+    exec { "copy easy-rsa to openvpn config folder ${name}":
```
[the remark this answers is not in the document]
NoodlesNZ: Not sure what you mean about this. It's just adding a -L to the cp command to dereference symlinks (as this was causing problems). Original code here: https://github.com/voxpupuli/puppet-openvpn/blob/master/manifests/ca.pp#L127-L131

templates/vars-30.erb
NoodlesNZ: I'm struggling to make the epp template work with the parameters from the defined type, `openvpn::ca`. Any guidance here? Should these be passed in as a parameter on the `epp()` function?

jameskirsop:
Any news on when this PR might get merged?

NoodlesNZ:
At the moment it's stuck with a broken EPP template, so can't be merged until that is fixed. See https://github.com/voxpupuli/puppet-openvpn/pull/278#discussion_r191621363

manifests/ca.pp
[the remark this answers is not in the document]
bastelfreak: the [file](https://puppet.com/docs/puppet/5.5/types/file.html) resource can copy local directories and files, I don't see why this needs to be an exec resource.

templates/vars-30.erb
Dan33l: Perhaps read this link will permit to understand the global context : 
https://tickets.puppetlabs.com/browse/PUP-7276

A possible way is like into `puppetlabs-apt` module with defined type `apt::setting`.
An other way, as you suggested, is to pass a hash as the last argument of the `epp()` function.

manifests/ca.pp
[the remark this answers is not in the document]
Dan33l: Some details : the [source attribute](https://puppet.com/docs/puppet/5.5/types/file.html#file-attribute-source) of file type with `file:` URI.

Dan33l:
This PR is close since #292 is merged.

PR #279, opened 2018-05-16
Do not require openvpn::client for client specific configs

### LINKED ISSUE #229: Client specific configs with extca enabled
`clinet_specific_config` type requires `openvpn::client` which fails when server has `extca_enabled => true`. It would be great if we can remove/workaround that dependency and use `clinet_specific_config` type with extca enabled.

spacedog:
There is no need to require openvpn::client type as client configs
don't rely on  any of that type resources. It also not possible to use
client_specific_config if extca_enabled is set to true for a server due
to openvpn::client failure.

Fixes #229

Signed-off-by: Anton Baranov <abaranov@linuxfoundation.org>

### DIFF
--- a/manifests/client_specific_config.pp
+++ b/manifests/client_specific_config.pp
@@ -88,7 +88,6 @@ define openvpn::client_specific_config (
 ) {
 
   Openvpn::Server[$server]
-  -> Openvpn::Client[$name]
   -> Openvpn::Client_specific_config[$name]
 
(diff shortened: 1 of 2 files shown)


bastelfreak:
@spacedog thanks for the PR. Can you please rebase?

bastelfreak:
Hi. I'm going to close this PR due to inactivity. Please reopen and rebase it if you're still interested.

PR #284, opened 2018-06-25
Extending mail capabilities

antgel:
Emails openvpn configuration to specified user(s).

### DIFF
--- a/files/send_mail.py
+++ b/files/send_mail.py
@@ -1,0 +1,76 @@
+#!/usr/bin/python
+# Python code for sending mail with attachments
+ 
+# libraries to be imported
+import smtplib
+import argparse, os
+from email.mime.multipart import MIMEMultipart
+from email.mime.text import MIMEText
+from email.mime.base import MIMEBase
+from email import encoders
+
+def send_mail (
+	file_name,
+	receiver,
+	sender,
+):
+	SMTPSERVER = "localhost"
+
+	# instance of MIMEMultipart
+	msg = MIMEMultipart()
+ 
+	# storing the senders email address  
+	msg['From'] = sender
+ 
+	# storing the receivers email address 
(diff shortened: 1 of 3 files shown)


bastelfreak:
HI @antgel, thanks for this PR. I'm not sure if this is a good approach. Do we really need a python script to send out emails? Mail handling is a very very complicated topic because almost each environment is different.  It's maybe the best idea to realize mail handling with a dedicated module. I'm open for opinions from our @voxpupuli/collaborators here.

xaque208:
I'd think this sort of code belongs in the manifests for a given infrastructure instead of the module here.  In my opinion, this makes too many assumptions to be useful generally, and strays away from the core of the module's primary purpose.  The exec here could still subscribe the resources in the module, but not need to be included with this project.

wyardley:
Agree with @bastelfreak and @xaque208 here

antgel:
Hi all, thanks for the feedback. @shapi78 wrote this, because it fits an internal need. We wanted to send it upstream to give back - perhaps we should have consulted before hacking away. I can't see us having time to devote to a significant refactor, so should I close this?

Dan33l:
With PR as this :
  * to install python on node becomes mandatory . Python is not installed on supported OSes by default.
  * to receive mail is mandatory. I expect a switch true/false to define behaviour.
  * to use `localhost` as smtp relay is mandatory. I expect a parameter.

In my point of view,  3 locking points that should be resolved.

oranenj:
@antgel I am not opposed to merging this functionality with the python script if it's documented in the README that python is required and that you make the other smaller changes I noted. It can be changed to be more portable later if that's required, most likely without affecting existing users.

bastelfreak:
I still think that this, in the current implementation, is way to site-specific and should be handled in a profile.

PR #292, opened 2018-09-02
use new fact easyrsa to configure easyrsa 2 or 3

### LINKED ISSUE #269: Epel has upgraded `easy-rsa` to version 3.x and removed 2.x, breaking the module
N/A (At least epel for EL7)

## What are you seeing
`easy-rsa` 3.x has a completely different way of working to version 2.x and so the commands in the module fail to work (`pkitool` for example).
It seems the commands have been rolled into the `easyrsa`  script supplied, but the commands don't seem to quite map 1-2-1 with the original structure of 2.x.

Dan33l:

A rework of PR #278 

Fixes #269

### DIFF
--- a/lib/facter/easyrsa.rb
+++ b/lib/facter/easyrsa.rb
@@ -1,7 +1,6 @@
 Facter.add(:easyrsa) do
   confine kernel: 'Linux'
   setcode do
-    version = ''
     binaryv2 = ''
     binaryv3 = ''
     operatingsystem = Facter.value(:operatingsystem)
@@ -28,18 +27,22 @@ Facter.add(:easyrsa) do
       binaryv3 = '/usr/local/share/easy-rsa/easyrsa'
     end
 
-    if File.exist? binaryv2
-      data = Facter::Core::Execution.execute("#{binaryv2} --help")
-      version = '2.0' if data.gsub!(%r{pkitool 2.0}, '')
-    elsif File.exist? binaryv3
+    if File.exist? binaryv3
       data = Facter::Core::Execution.execute("#{binaryv3} --help")
       version = '3.0' if data.gsub!(%r{Easy-RSA 3 usage}, '')
+    elsif File.exist? binaryv2
+      data = Facter::Core::Execution.execute("#{binaryv2} --help")
+      version = '2.0' if data.gsub!(%r{pkitool 2.0}, '')
     elsif Facter::Util::Resolution.which('pkitool')
       data = Facter::Core::Execution.execute('pkitool --help')
       version = '2.0' if data.gsub!(%r{pkitool 2.0}, '')
     elsif Facter::Util::Resolution.which('easyrsa')
       data = Facter::Core::Execution.execute('easyrsa --help')
       version = '3.0' if data.gsub!(%r{Easy-RSA 3 usage}, '')
+    end
+    if !version.nil?
+    else
+      version = nil
     end
     version
   end
--- a/lib/facter/easyrsa.rb
+++ b/lib/facter/easyrsa.rb
@@ -1,0 +1,57 @@
+Facter.add(:easyrsa) do
+  confine :kernel => 'Linux'
+  setcode do
+    version = ''
+    binaryv2 = ''
+    binaryv3 = ''
+    operatingsystem = Facter.value(:operatingsystem)
+    operatingsystemrelease = Facter.value(:operatingsystemrelease)
+
+    case operatingsystem
+    when /RedHat|CentOS/
+      binaryv2 = '/usr/share/easy-rsa/2.0/pkitool'
+      binaryv3 = '/usr/share/easy-rsa/3/easyrsa'
+    when /Ubuntu|Debian/
+      case operatingsystemrelease
+      when /8|9|16.04|18.04/
+        binaryv2 = '/usr/share/easy-rsa/pkitool'
+        binaryv3 = '/usr/share/easy-rsa/easyrsa'
+      else
+        binaryv2 = '/usr/share/doc/openvpn/examples/easy-rsa/2.0/pkitool'
+        binaryv3 = '/usr/share/doc/openvpn/examples/easy-rsa/3.0/easyrsa'
+      end
+    when /Archlinux/
+      binaryv2 = '/usr/share/easy-rsa/2.0/pkitool'
+      binaryv3 = '/usr/share/easy-rsa/3/easyrsa'
+    when /Amazon/
+      binaryv2 = '/usr/share/easy-rsa/2.0/pkitool'
+      binaryv3 = '/usr/share/easy-rsa/3/easyrsa'
+    when /FreeBSD/
+      binaryv2 = '/usr/local/share/easy-rsa/pkitool'
+      binaryv3 = '/usr/local/share/easy-rsa/easyrsa'
+    end
+
+    if File.exist? binaryv2
+      Open3.popen2("#{binaryv2} --help") do |stdin, stdout, status_thread|
+        stdout.each_line do |line|
+          if line.include?('pkitool 2.0')
+            version = '2.0'
+            break
+          end
+        end
+      end
+      version
+    elsif File.exist? binaryv3
+      Open3.popen2("#{binaryv3} --help") do |stdin, stdout, status_thread|
+        stdout.each_line do |line|
+          if line.include?('Easy-RSA 3 usage and overview')
+            version = '3.0'
+            break
+          end
+        end
+      end
+      version
+    end
+  end
+end
+
--- a/lib/facter/easyrsa.rb
+++ b/lib/facter/easyrsa.rb
@@ -1,5 +1,5 @@
 Facter.add(:easyrsa) do
-  confine :kernel => 'Linux'
+  confine kernel: 'Linux'
   setcode do
     version = ''
     binaryv2 = ''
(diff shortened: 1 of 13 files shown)


lib/facter/easyrsa.rb
```diff
@@ -0,0 +1,57 @@
...
+      when /8|9|16.04|18.04/
+        binaryv2 = '/usr/share/easy-rsa/pkitool'
+        binaryv3 = '/usr/share/easy-rsa/easyrsa'
+      else
+        binaryv2 = '/usr/share/doc/openvpn/examples/easy-rsa/2.0/pkitool'
+        binaryv3 = '/usr/share/doc/openvpn/examples/easy-rsa/3.0/easyrsa'
+      end
+    when /Archlinux/
+      binaryv2 = '/usr/share/easy-rsa/2.0/pkitool'
+      binaryv3 = '/usr/share/easy-rsa/3/easyrsa'
```
bastelfreak: `/usr/bin/easyrsa` is the correct path

lib/facter/easyrsa.rb
```diff
@@ -0,0 +1,57 @@
...
+    when /Amazon/
+      binaryv2 = '/usr/share/easy-rsa/2.0/pkitool'
+      binaryv3 = '/usr/share/easy-rsa/3/easyrsa'
+    when /FreeBSD/
+      binaryv2 = '/usr/local/share/easy-rsa/pkitool'
+      binaryv3 = '/usr/local/share/easy-rsa/easyrsa'
+    end
+
+    if File.exist? binaryv2
+      Open3.popen2("#{binaryv2} --help") do |stdin, stdout, status_thread|
```
bastelfreak: Puppet has some tooling to check if a binary is within `$PATH` and can also execute it:
https://github.com/voxpupuli/puppet-rabbitmq/blob/d6a216f985e421e81ac3e66dcfcd59faaada9da1/lib/facter/erl_ssl_path.rb#L7-L8

lib/facter/easyrsa.rb
```diff
@@ -0,0 +1,57 @@
...
+        stdout.each_line do |line|
+          if line.include?('pkitool 2.0')
+            version = '2.0'
+            break
+          end
+        end
+      end
+      version
+    elsif File.exist? binaryv3
```
bastelfreak: if both files aren't available, we should check if it is within the `$PATH`

manifests/ca.pp
```diff
@@ -124,86 +131,158 @@
...
-    require => Exec["copy easy-rsa to openvpn config folder ${name}"],
+    require => File["${etc_directory}/openvpn/${name}/easy-rsa"],
   }
 
-  file { "${etc_directory}/openvpn/${name}/easy-rsa/vars":
-    ensure  => file,
-    mode    => '0550',
-    content => template('openvpn/vars.erb'),
-    require => Exec["copy easy-rsa to openvpn config folder ${name}"],
+  if $easyrsa_version == '2.0' {
```
bastelfreak: should we use[ versioncmp()](https://puppet.com/docs/puppet/5.5/function.html#versioncmp) here?

manifests/ca.pp
```diff
@@ -124,86 +131,158 @@
...
+      ensure  => file,
+      mode    => '0550',
+      content => template('openvpn/vars.erb'),
+      require => File["${etc_directory}/openvpn/${name}/easy-rsa"],
+    }
+  } else {
+    file { "${etc_directory}/openvpn/${name}/easy-rsa/vars":
+      ensure  => file,
+      mode    => '0550',
+      content => epp('openvpn/vars-30.epp',
```
bastelfreak: yey epp with params \o/

manifests/params.pp
```diff
@@ -142,11 +151,13 @@
...
       $easyrsa_source      = '/usr/local/share/easy-rsa'
+      $default_easyrsa_ver = '2.0'
       $namespecific_rclink = true
       $systemd             = false
     }
     default: {
       fail("Not supported OS family ${::osfamily}")
     }
   }
+
```
bastelfreak: is this newline on purpose?

spec/spec_helper_acceptance.rb
```diff
@@ -0,0 +1,36 @@
...
+install_module_dependencies
+
+RSpec.configure do |c|
+
+  # Configure all nodes in nodeset
+  c.before :suite do
+    hosts.each do |host|
+      case host[:platform]
+      when /debian-7-amd64|debian-8-amd64|ubuntu-16.04-amd64/
+        on host, puppet('module', 'install', 'puppetlabs-apt')
```
bastelfreak: you can do something like this:
```ruby
install_module_from_forge('puppetlabs-apt', '>= 4.1.0 < 7.0.0') if fact('os.family') == 'Debian'
```

spec/spec_helper_acceptance.rb
```diff
@@ -0,0 +1,36 @@
...
+        pp = <<-EOS
+        package { 'netcat-openbsd' :
```
bastelfreak: you can install packages like this:
```ruby
install_package(host, 'netcat-openbsd') if fact('os.release.major')
```

templates/server.erb
```diff
@@ -26,11 +26,21 @@ dh <%= @extca_dh_file %>
 crl-verify <%= @extca_ca_crl_file %>
 <%   end -%>
 <% else -%>
+<%- if scope.lookupvar('::openvpn::params::easyrsa_ver') == '2.0' -%>
```
bastelfreak: please use ```@openvpn::params::easyrsa_ver``` to access variables (or if you are interested you can turn this into an epp template, but that sould probably done in another PR).

templates/vars-30.epp
```diff
@@ -0,0 +1,80 @@
+# easy-rsa parameter settings
```
bastelfreak: does this actually need to be a template? Can't we use the files distributed by the operatingsystems?

lib/facter/easyrsa.rb
Dan33l: For any version of easyrsa ?

manifests/ca.pp
Dan33l: I tryed to KISS. To use the function `versioncmp()` is some time usefull. But here, in my opinion it will not provide better result. But i'll do as you prefer. Tell me.

manifests/params.pp
Dan33l: typo.

templates/server.erb
Dan33l: I need to modify this, so it will be handled.

templates/vars-30.epp
Dan33l: For me, template usage is still relevant because the file `vars` need to be customized for each `openvpn` tunnel.

If you are thinking about something else. Please tell me more.

spec/spec_helper_acceptance.rb
Dan33l: `netcat` was used in a previous version of tests. But now it is not. I'll remove this.

spec/spec_helper_acceptance.rb
Dan33l: I'll use function for `epel` module.
And this  comment makes me realize that `Debian` family OSes does not need special apt sources.

lib/facter/easyrsa.rb
bastelfreak: for v3. v2 does not exist on Arch Linux.

templates/vars-30.epp
bastelfreak: ah okay. Than it makes sense.

lib/facter/easyrsa.rb
Dan33l: Nice. Will be used.

lib/facter/easyrsa.rb
Dan33l: I agree. Will be handled.

Dan33l:
Comments from @bastelfreak's first review are handled.

If you run acceptance tests you will show that 2 runs are needed to make all actions.
This is due to : 
  * `exec` creates certificates for server and clients in `keys` directory as `root`. and so files are owned by `root.root`.
  * a `file` resource is used to make a copy fom `easyrsa` path to `openvpn` instance.
  * some default settings are defined https://github.com/Dan33l/puppet-openvpn/blob/add-easyrsa-3/manifests/ca.pp#L116

I tried to add `notify` to `exec` by i meet a dependency cycle.
Advice to fix this is welcome.

lib/facter/easyrsa.rb
```diff
@@ -0,0 +1,46 @@
...
+      end
+    when %r{Amazon}
+      binaryv2 = '/usr/share/easy-rsa/2.0/pkitool'
+      binaryv3 = '/usr/share/easy-rsa/3/easyrsa'
+    when %r{FreeBSD}
+      binaryv2 = '/usr/local/share/easy-rsa/pkitool'
+      binaryv3 = '/usr/local/share/easy-rsa/easyrsa'
+    end
+
+    if File.exist? binaryv2
```
bastelfreak: should we flip the logic here? v2 wil be used, even if v3 is available. Shouldn't we always use the latest available version?

manifests/params.pp
```diff
@@ -142,11 +151,17 @@
...
       $namespecific_rclink = true
       $systemd             = false
     }
     default: {
       fail("Not supported OS family ${::osfamily}")
     }
   }
+  $easyrsa_version = $facts['easyrsa'] ? {
+    undef   => $default_easyrsa_ver,
+    ''      => $default_easyrsa_ver,
```
bastelfreak: the fact should be reworked to only be undef or return a string.

manifests/client.pp
```diff
@@ -347,6 +374,13 @@
     ],
   }
 
+  file { "${etc_directory}/openvpn/${server}/download-configs/${name}.tar.gz":
+    ensure  => present,
+    replace => 'no',
+    content => "not yet created\n",
```
bastelfreak: why is there a content attribute for a .tar.gz file?

spec/spec_helper_acceptance.rb
```diff
@@ -0,0 +1,20 @@
+require 'beaker-rspec'
+require 'beaker-puppet'
+require 'beaker/puppet_install_helper'
+require 'beaker/module_install_helper'
+
+run_puppet_install_helper unless ENV['BEAKER_provision'] == 'no'
```
bastelfreak: do the acceptance tests also work with docker? If so, we should enable them for travis.

lib/facter/easyrsa.rb
Dan33l: Ooooooh yes !

manifests/params.pp
Dan33l: done

manifests/client.pp
Dan33l: It works also without. I wanted to have the possibility to know if something was wrong with content of file.

content removed.

spec/spec_helper_acceptance.rb
Dan33l: Yes it works with docker.

bastelfreak:
Thanks @Dan33l !

PR #294, opened 2018-09-10
move concat version_requirement to >= 3.0.0 < 6.0.0

Dan33l:

To fix broken modules dependencies management . 
With current settings in `metadata.json` a fresh install get the following error message : 
```
Error: Could not install module 'puppetlabs-concat' (???)
  	  No version of 'puppetlabs-concat' can satisfy all dependencies
  	    Use `puppet module install --ignore-dependencies` to install only this module
```

It will permit to merge the PR #292

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -57,7 +57,7 @@
     },
     {
       "name": "puppetlabs/concat",
-      "version_requirement": ">= 3.0.0 < 5.0.0"
+      "version_requirement": ">= 3.0.0 < 6.0.0"
     }
   ],
   "tags": [


metadata.json
```diff
@@ -57,7 +57,7 @@
     },
     {
       "name": "puppetlabs/concat",
-      "version_requirement": ">= 3.0.0 < 5.0.0"
+      "version_requirement": ">= 5.0.0 < 6.0.0"
```
alexjfisher: Is 5.0.0 really a minimum requirement?

Dan33l: If you try `"version_requirement": ">= 4.0.0 < 5.0.0"`, you will get also an error about unsatisfied dependencies.

bastelfreak: this should probably be ```">= 3.0.0 < 6.0.0"```

dhollinger: @Dan33l have you tried `"version_requirement": ">= 4.0.0 < 6.0.0"`? Let's not bump the minimum version until we have to

Dan33l: @dhollinger i didn't noted all my tests. 
The range you proposed looks to work well.
I push this.

Dan33l:
The range `>= 3.0.0 < 6.0.0` looks to work also.

PR #296, opened 2018-09-14
update supported OSes in params.pp

### LINKED ISSUE #291: debian 7 support broken

operating system debian 7

```
      openvpn::server { 'test_openvpn_server':
        country      => 'FR',
        province     => 'Occitanie',
        city         => 'France',
        organization => 'FOO',
        email        => 'bar@foo.org',
        server       => '10.0.0.0 255.255.255.0',
        local        => '',
      }
```

## What are you seeing

When this module is used on debian 7, the current settings about cipher with a so old operating system, does not work. The service `openvpn` does not start

process `openvpn` started.

## Output log

The start of openvpn server log the following : 
`Problem with cipher list: TLS-DHE-RSA-WITH-AES-256-GCM-SHA384:TLS-DHE-RSA-WITH-AES-256-CBC-SHA256:TLS-DHE-RSA-WITH-AES-128-GCM-SHA256:TLS-DHE-RSA-WITH-AES-128-CBC-SHA256: error:1410D0B9:SSL routines:SSL_CTX_set_cipher_list:no cipher match`

If removing debian 7 support is considered as possible option,  i can provide a PR.

Dan33l:

update `params.pp` to be up to date with current `metadata.json`

Fixes #291

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -16,98 +16,71 @@
 #
 class openvpn::params {
 
-  case $::osfamily {
-    'RedHat': {
+  case $facts['os']['family'] {
+    'RedHat': { # RedHat/CentOS
       $etc_directory       = '/etc'
       $root_group          = 'root'
       $group               = 'nobody'
       $link_openssl_cnf    = true
       $pam_module_path     = '/usr/lib64/openvpn/plugin/lib/openvpn-auth-pam.so'
       $namespecific_rclink = false
+      $default_easyrsa_ver = '3.0'
+      $easyrsa_source      = '/usr/share/easy-rsa/3'
 
-      # Redhat/Centos >= 7.0
-      if(versioncmp($::operatingsystemrelease, '7.0') >= 0) and $::operatingsystem != 'Amazon' {
-        $additional_packages = ['easy-rsa']
-        $ldap_auth_plugin_location = undef
-        $systemd = true
-        $easyrsa_source      = '/usr/share/easy-rsa/3'
-        $default_easyrsa_ver = '3.0'
-      # Redhat/Centos == 6.0
-      } elsif(versioncmp($::operatingsystemrelease, '6.0') >= 0) and $::operatingsystem != 'Amazon' {
-        $additional_packages = ['easy-rsa','openvpn-auth-ldap']
-        $ldap_auth_plugin_location = '/usr/lib64/openvpn/plugin/lib/openvpn-auth-ldap.so'
-        $systemd = false
-        $easyrsa_source      = '/usr/share/easy-rsa/3'
-        $default_easyrsa_ver = '3.0'
-      # Redhat/Centos < 6.0
-      } else {
-        $additional_packages = ['easy-rsa']
-        $ldap_auth_plugin_location = undef
-        $systemd = false
-        $easyrsa_source      = '/usr/share/easy-rsa/2.0'
-        $default_easyrsa_ver = '2.0'
+      case $facts['os']['release']['major'] {
+        '7': {
+          $additional_packages = ['easy-rsa']
+          $ldap_auth_plugin_location = undef
+          $systemd = true
+        }
+        '6': {
+          $additional_packages = ['easy-rsa','openvpn-auth-ldap']
+          $ldap_auth_plugin_location = '/usr/lib64/openvpn/plugin/lib/openvpn-auth-ldap.so'
+          $systemd = false
+        }
+        default: {
+          fail("unsupported OS ${facts['os']['name']} ${facts['os']['release']['major']}")
+        }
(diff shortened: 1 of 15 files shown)


manifests/params.pp
```diff
@@ -16,98 +16,71 @@
 #
 class openvpn::params {
 
-  case $::osfamily {
-    'RedHat': {
+  case $facts['os']['family'] {
```
bastelfreak: structured facts \o/

manifests/params.pp
```diff
@@ -16,98 +16,71 @@
...
+      case $facts['os']['release']['major'] {
+        '7': {
+          $additional_packages = ['easy-rsa']
+          $ldap_auth_plugin_location = undef
+          $systemd = true
+        }
+        '6': {
+          $additional_packages = ['easy-rsa','openvpn-auth-ldap']
+          $ldap_auth_plugin_location = '/usr/lib64/openvpn/plugin/lib/openvpn-auth-ldap.so'
+          $systemd = false
```
bastelfreak: suggestion for future enhancements: ```$systemd``` can be set to ```true``` if the fact ```[service_provider```](https://github.com/puppetlabs/puppetlabs-stdlib#service_provider) returns ```systemd```

spec/classes/openvpn_config_spec.rb
```diff
@@ -4,9 +4,9 @@
   context 'on Debian based machines' do
     let(:facts) do
       {
-        osfamily: 'Debian',
-        operatingsystem: 'Debian',
-        operatingsystemrelease: '7',
+        os: { 'family' => 'Debian' },
```
bastelfreak: this should be migrated to rspec-puppet-facts, same applies for the other spec files.

Dan33l: You're right, but the tag `needs-work` was not useless.
It did it.

spec/classes/openvpn_config_spec.rb
```diff
@@ -1,37 +1,41 @@
...
-        operatingsystem: 'Debian',
-        operatingsystemrelease: '7',
-        concat_basedir: '/var/lib/puppet/concat'
-      }
-    end
+  on_supported_os.each do |os, facts|
+    context "on #{os}" do
+      let(:facts) do
+        facts.merge(
+          concat_basedir: '/var/lib/puppet/concat'
```
bastelfreak: Did you test if this is required? We have mocked facts here: https://github.com/voxpupuli/puppet-openvpn/blob/9f3abf684f9d817e67a56469c15ea4028eae34d7/spec/default_facts.yml#L11

And use them here:
https://github.com/voxpupuli/puppet-openvpn/blob/9f3abf684f9d817e67a56469c15ea4028eae34d7/spec/spec_helper.rb#L25-L35

bastelfreak:
If I see this correctly this PR doesnt add support for new operating systems, right? Otherwise the list in the metadata.json needs to be updated as well.

bastelfreak:
I created a little patch that you might want to look into:

```patch
diff --git a/spec/classes/openvpn_config_spec.rb b/spec/classes/openvpn_config_spec.rb
index 10b4e73..d33e35d 100644
--- a/spec/classes/openvpn_config_spec.rb
+++ b/spec/classes/openvpn_config_spec.rb
@@ -4,12 +4,12 @@ describe 'openvpn::config', type: :class do
   on_supported_os.each do |os, facts|
     context "on #{os}" do
       let(:facts) do
-        facts.merge(
-          concat_basedir: '/var/lib/puppet/concat'
-        )
+        facts
       end
 
-      case facts[:os][:family]
+      it { is_expected.to compile.with_all_deps }
+
+      case facts[:os]['family']
       when 'Debian'
         context 'on Debian based machines' do
           it { is_expected.to contain_concat('/etc/default/openvpn') }
```

A couple of things:
* I tested this, we don't need to mock concat_basedir
* in the facts hash, only the first key is a symbol, all further nested keys are strings
* It's always good to check if the manifest compiles into a catalog

Dan33l:
> If I see this correctly this PR doesnt add support for new operating systems, right? Otherwise the list in the metadata.json needs to be updated as well.

Exactly . No new OS.

manifests/server.pp
```diff
@@ -454,7 +454,7 @@
   Optional[String] $group                                           = undef,
   Boolean $ipp                                                      = false,
   Boolean $duplicate_cn                                             = false,
-  String $local                                                     = $::ipaddress_eth0,
+  String $local                                                     = $facts['ipaddress_eth0'],
```
bastelfreak: facts hash \o/

bastelfreak:
thanks for this awesome PR!

Issue #172, opened 2015-08-28
openvpn-auth-ldap package is available in epel 6 repos


earsdown:
Hey,
Doesn't affect me, but it looks like openvpn-auth-ldap is available in the EPEL 6 repos for redhat/centos 6.x machines.
Cheers

luxflux:
Thanks for the heads up. I will leave this open, if anyone wants to implement this, I am open to PRs.

luxflux:
Also see [this discussion](https://github.com/luxflux/puppet-openvpn/commit/bae0c0e193306c3f03b81d2ed61e710dee4d6a68#commitcomment-13612143).

ciobancai:
The following package for "openvpn-auth-ldap" is not require to be installed on RHEL7. The plugin is already include with the latest openvpn package and can be found at this location: /usr/lib64/openvpn/plugins/openvpn-plugin-auth-pam.so

I'm using openvpn-2.3.10-1.el7.x86_64 and worked OK

WetHippie:
I'm stuck with this as well (Centos 7 here).  Even manually installing the plugin off the command line does not allow the installation to proceed.

WetHippie:
By using the PAM module, my openvpn will allow access directly without needing other modules. I already have a global PAM -> LDAP configuration running, so this just makes use of that.  On Centos7 the path and filename are different, so I've just pushed in conditional path selection based on version.

ekohl:
Looks like the data is set for EL6
https://github.com/voxpupuli/puppet-openvpn/blob/258ac2b5834726cf2ca4df70caa3888403586a89/data/family/RedHat/6.yaml#L1-L2

Closing this. If a similar fix is needed for EL7 then please open a new issue.

PR #300, opened 2018-10-15
fix for issue #297, adding the path for the ldap auth plugin on CentOS 7

### LINKED ISSUE #297: openvpn-auth-ldap.so plugin not being loaded in CentOS 7
I am deploying an openvpn server on CentOS 7, and openvpn fails to start due to an invalid config. I've traced it to this line (https://github.com/voxpupuli/puppet-openvpn/blob/master/manifests/params.pp#L33) unsetting the lib file. This writes out an invalid plugin line in the config causing the following error to appear.

`
Thu Oct 11 10:49:30 2018 us=958804 PLUGIN_INIT: could not load plugin shared object /etc/openvpn/vpnserver/auth/ldap.conf: /etc/openvpn/vpnserver/auth/ldap.conf: invalid ELF header: No such file or directory (errno=2)
`

The config looks like this:
`
plugin  "/etc/openvpn/vpnserver/auth/ldap.conf"
`

when it should look like this:
`
plugin /usr/lib64/openvpn/plugin/lib/openvpn-auth-ldap.so "/etc/openvpn/vpnserver/auth/ldap.conf"
`

Error is in the current commit, as well as the module on puppetforge which I'm currently trying to use.

kindred:
A fix for the malformed config file on CentOS 7 when using the LDAP auth plugin.

Fixes #297


manifests/params.pp
```diff
@@ -30,7 +30,7 @@
       case $facts['os']['release']['major'] {
         '7': {
           $additional_packages = ['easy-rsa']
```
alexjfisher: Should this now also be `$additional_packages = ['easy-rsa','openvpn-auth-ldap']` for EL7 as well as EL6?
(If these parameters are now the same on EL6 and EL7, they can also be moved out of this case statement).

alexjfisher: Did EL7 packages become available some time *after* https://github.com/voxpupuli/puppet-openvpn/pull/215 ?

Dan33l:
Looks related with issue #172

alexjfisher:
openvpn-auth-ldap is definately available in [epel 7](https://dl.fedoraproject.org/pub/epel/7/x86_64/Packages/o/openvpn-auth-ldap-2.0.3-15.el7.x86_64.rpm) these days.  No idea since when though.

manifests/params.pp
kindred: It looks like the openvpn-auth-ldap package is now available in the repo for EL7. Will amend the PR.

Dan33l:
Oups ? why closed ?

PR #301, opened 2018-10-15
FreeBSD: change additional_packages to easy-rsa2

olevole:
the default easy-rsa package in FreeBSD gets easy-rsa 3.0.5
which for some reason does not work correctly in puppet-openvpn
under FreeBSD platform ( of course with changed $default_easyrsa_ver )
Therefore, temporarily specify the correct easy-rsa 2x package name
for FreeBSD: https://www.freshports.org/security/easy-rsa2/

I will try to look soon for easy-rsa 3.0.5 and puppet-openvpn error
with FreeBSD platform.

Tested on:
  FreeBSD 11
  FreeBSD 12
  ( Puppet client: puppet5-5.5.6, Puppet server: 5.3.5 )

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -103,7 +103,7 @@ class openvpn::params {
       $group               = 'nogroup'
       $link_openssl_cnf    = true
       $pam_module_path     = '/usr/local/lib/openvpn/openvpn-auth-pam.so'
-      $additional_packages = ['easy-rsa']
+      $additional_packages = ['easy-rsa2']
       $easyrsa_source      = '/usr/local/share/easy-rsa'
       $default_easyrsa_ver = '2.0'
       $namespecific_rclink = true


Dan33l:
So, what i am understanding is that since FreeBSD provides easyrsa version 3 : 
 * package `easyrsa` is for version 3
 * package `easyrsa2` is for version 2

And because of problems you encountered with this module and `easyrsa3` or FreeBSD, you PR does not modify `$default_easyrsa_ver`.

We do not have acceptance with this OS. I prefer to be sure to understand.

olevole:
@Dan33l Yes, that's right,  easyrsa3 is another story. By this PR we will only fix the puppet-openvpn module for FreeBSD (_with easyrsa2_).

Dan33l:
Feel free to propose a PR about easyrsa version 3 on FreeBSD.

PR #317, opened 2018-12-06
Set DH timeout to accommodate low performance CPU

### LINKED ISSUE #316: Timeout when generating Diffie-Hellman  parameters on a low-performance CPU

- Puppet: 4.8.2
- Ruby: 2.3.3p222
- Distribution: Respbian
- Module version: 7.4.0

Try to configure OpenVPN with the default setup on a Raspberry Pi B+.

```
  openvpn::server { 'winterthur':
    country      => 'CH',
    province     => 'ZH',
    city         => 'Winterthur',
    organization => 'example.org',
    email        => 'root@example.org',
    server       => '10.200.200.0 255.255.255.0',
  }
```

## What are you seeing

The Puppet script times out while executing the rule `generate dh param` for the command `./build-dh`

I would expect the server to be configured without an error.

## Any additional information you'd like to impart
On a Raspberry Pi B+ the Diffie-Hellman  parameter generation for a 2048 key was measured to take 229m45.379s wall clock time.

I submitting a PR that fixes the problem.

dspinellis:
#### Set DH timeout to accommodate computers with a low performance CPU
On a Raspberry Pi B+ the Diffie-Hellman  parameter generation for
a 2048 key was measured to take 229m45.379s wall clock time.
Set the respective command's timeout 50% above this, to prevent
the corresponding rule from timing out.

Fixes #316

### DIFF
--- a/manifests/ca.pp
+++ b/manifests/ca.pp
@@ -92,7 +92,7 @@ define openvpn::ca (
 
       exec { "generate dh param ${name}":
         command  => '. ./vars && ./clean-all && ./build-dh',
-        timeout  => 1800,
+        timeout  => 20000,
         cwd      => "${etc_directory}/openvpn/${name}/easy-rsa",
         creates  => "${etc_directory}/openvpn/${name}/easy-rsa/keys/dh${ssl_key_size}.pem",
         provider => 'shell',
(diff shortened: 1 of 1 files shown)


manifests/ca.pp
```diff
@@ -92,7 +92,7 @@
 
       exec { "generate dh param ${name}":
         command  => '. ./vars && ./clean-all && ./build-dh',
-        timeout  => 1800,
+        timeout  => 20000,
```
Dan33l: I am seeing that is EasyRSA version is 3, then `timeout` is not present.
Can you add same `timeout` in [the other exec generate dh](https://github.com/voxpupuli/puppet-openvpn/blob/52fd0d61dce2539c775420348cf4ba5ea2a5c518/manifests/ca.pp#L167)  also ?

dspinellis:
Good point; I added it.

Dan33l:
@dspinellis please squash your commits and the PR will be merged when CI will be successfully finished.

PR #319, opened 2018-12-07
Remove conditional statements

### LINKED ISSUE #318: crl auto renewal broken with easyrsa 3.0
If `EASY_REQ_CN` and `EASYRSA_REQ_OU` aren't passed in the `openvpn::server` definition, and are blank (which is the default for those params in the `ca.pp` class definition, the vars template doesn't allow for setting the related environment vars.

When `crl_auto_renew => true`, the missing environment vars causes the auto CRL renew process to fail (silently) and the CRL expires when it shouldn't be doing so.

If these two variables are indeed manually set to empty strings, the renewal process works without a hitch. The 3.0 template therefore shouldn't have the conditional clauses around these two params, and instead always write them to the `easy-rsa/vars` file regardless of their contents.

jameskirsop:
This PR removes the conditional statements around two environment variables so that they are always set. In doing so it allows the auto crl renewal process to succeed if either of the two options are missing in the `openvpn::server` declaration.

Fixes #318

### DIFF
--- a/templates/vars-30.epp
+++ b/templates/vars-30.epp
@@ -72,9 +72,5 @@ export EASYRSA_REQ_PROVINCE="<%= $province %>"
 export EASYRSA_REQ_CITY="<%= $city %>"
 export EASYRSA_REQ_ORG="<%= $organization %>"
 export EASYRSA_REQ_EMAIL="<%= $email %>"
-<% if $key_cn != '' { -%>
 export EASYRSA_REQ_CN="<%= $key_cn %>"
-<% } -%>
-<% if $key_ou != '' { -%>
 export EASYRSA_REQ_OU="<%= $key_ou %>"
-<% } -%>


Dan33l:
Hi @jameskirsop i don't thing the fix proposed is the good one. All acceptance tests on nodes running EasyRSA v3 are failing.

IMO, the issue is that the command used in resource `openvpn::server` does not make the difference if the node use EasyRSA v2 or v3. 
With EasyRSA v3, i tested the following command  and it works like a charme: 
`. ./vars && EASYRSA_REQ_CN='' EASYRSA_REQ_OU='' openssl ca -gfncrl -out /etc/openvpn/test_openvpn_server/crl.pem -config /etc/openvpn/test_openvpn_server/easy-rsa/openssl.cnf`

So the fix should be, with custom fact `easyrsa`, to take care of the EasyRSA version installed and then trigger the corresponding command.

Can you modify the PR in this way ?

Dan33l:
If you need help fill free to ask. You can also join #voxpupuli on IRC and slack.

Dan33l:
The issue is now closed by PR #321

PR #324, opened 2019-01-05
Use stdlib functions for hash key discovery

towo:

When fiddling around with server names in the configuration, you'd have
to do a clean sweep of the whole configuration to ensure that there are
no problems with the fact existence and the ordering.

Using dig (and has_key for compatibility reasons to not change anything
front-facing), we can avoid this problem.

### DIFF
--- a/manifests/deploy/export.pp
+++ b/manifests/deploy/export.pp
@@ -18,8 +18,8 @@ define openvpn::deploy::export (
   -> Openvpn::Client[$name]
   -> Openvpn::Deploy::Export[$name]
 
-  if $facts['openvpn'] {
-    if $facts['openvpn'][$server][$name] {
+  if has_key($facts, 'openvpn') {
+    if dig44($facts, ['openvpn', $server, $name], false) {
       $data = $facts['openvpn'][$server][$name]
 
       @@file { "exported-${server}-${name}-config":
--- a/manifests/deploy/export.pp
+++ b/manifests/deploy/export.pp
@@ -18,63 +18,59 @@ define openvpn::deploy::export (
   -> Openvpn::Client[$name]
   -> Openvpn::Deploy::Export[$name]
 
-  if has_key($facts, 'openvpn') {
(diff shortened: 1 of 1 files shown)


manifests/deploy/export.pp
```diff
@@ -18,8 +18,8 @@
   -> Openvpn::Client[$name]
   -> Openvpn::Deploy::Export[$name]
 
-  if $facts['openvpn'] {
-    if $facts['openvpn'][$server][$name] {
+  if has_key($facts, 'openvpn') {
+    if dig44($facts, ['openvpn', $server, $name], false) {
```
bastelfreak: couldn't we just use:
```puppet
if fact("openvpn.${server}.${name}") {
```

towo: Would also work out, yeah, dig is just the pattern to use when you can't assert a hash structure in general… of course, `fact` in this specific instance is a more specific way around the whole issue.

(I left the `has_key` bit in as I didn't want to change the error cases as presented so far: fail if the whole fact doesn't exist, otherwise just consider working or not.)

towo:
Oh no, the builds are getting stuck again.

bastelfreak:
bundler 2.0 is causing this. We already added a fix to our modulesync_config, but haven't rolled it out yet.

manifests/deploy/export.pp
```diff
@@ -18,8 +18,8 @@
   -> Openvpn::Client[$name]
   -> Openvpn::Deploy::Export[$name]
 
-  if $facts['openvpn'] {
-    if $facts['openvpn'][$server][$name] {
+  if has_key($facts, 'openvpn') {
```
alexjfisher: `has_key` is deprecated. https://github.com/puppetlabs/puppetlabs-stdlib#has_key
You'd normally replace it with `in`, but in this case (as @bastelfreak suggests), I think the `fact` function is probably best.

bastelfreak:
thanks!

PR #327, opened 2019-02-02
Allow 4 and 6 suffix inside proto to limit ip4 or ip6 connection only.

jkroepke:
Since 8.0.0 it's not possible to choose `tcp4` or `tcp6` as `proto` for an openvpn server.

N/A

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -136,7 +136,7 @@ define openvpn::server (
   Variant[Boolean, String] $logfile                                 = false,
   String $port                                                      = '1194',
   Optional[String] $portshare                                       = undef,
-  Enum['tcp', 'udp'] $proto                                         = 'tcp',
+  Enum['tcp', 'tcp4', 'tcp6', 'udp', 'udp4', 'udp6'] $proto         = 'tcp',
   Enum['1', '2', '3', ''] $status_version                           = '',
   String $status_log                                                = "/var/log/openvpn/${name}-status.log",
   String $server                                                    = '',
--- a/spec/defines/openvpn_server_spec.rb
+++ b/spec/defines/openvpn_server_spec.rb
@@ -110,6 +110,81 @@ describe 'openvpn::server' do
         it { is_expected.to contain_file('/etc/openvpn/test_server.conf').with_content(%r{^rcvbuf\s+393215$}) }
       end
 
+      context 'when using udp4' do
+        let(:params) do
+          {
+            'country'       => 'CO',
+            'province'      => 'ST',
+            'city'          => 'Some City',
+            'organization'  => 'example.org',
+            'email'         => 'testemail@example.org',
+            'proto'         => 'udp4'
+          }
+        end
+
+        it { is_expected.to contain_file('/etc/openvpn/test_server.conf').with_content(%r{^proto\s+udp4$}) }
+      end
+
+      context 'when using udp6' do
+        let(:params) do
+          {
+            'country'       => 'CO',
+            'province'      => 'ST',
+            'city'          => 'Some City',
+            'organization'  => 'example.org',
+            'email'         => 'testemail@example.org',
+            'proto'         => 'udp6'
+          }
+        end
+
+        it { is_expected.to contain_file('/etc/openvpn/test_server.conf').with_content(%r{^proto\s+udp6$}) }
+      end
+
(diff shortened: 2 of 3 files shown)


manifests/server.pp
```diff
@@ -136,7 +136,7 @@
   Variant[Boolean, String] $logfile                                 = false,
   String $port                                                      = '1194',
   Optional[String] $portshare                                       = undef,
-  Enum['tcp', 'udp'] $proto                                         = 'tcp',
+  Enum['tcp', 'tcp4', 'tcp6', 'udp'] $proto                         = 'tcp',
```
bastelfreak: what about udp? Is there also an `udp4` and `udp6`?

jkroepke: Yes. I forgot that. Thanks

jkroepke:
Let me add some tests before merge.

templates/server.erb
```diff
@@ -46,7 +46,7 @@ dh <%= @etc_directory -%>/openvpn/<%= @ca_name %>/keys/dh.pem
 crl-verify <%= @etc_directory -%>/openvpn/<%= @ca_name %>/crl.pem
 <%   end -%>
 <% end -%>
-<% if @proto == 'tcp' -%>
+<% if @proto != 'udp' -%>
```
bastelfreak: is this still correct? Or should it be something like this (untested):
```suggestion
<% unless ['udp', 'udp4', 'udp6'].include?(@proto) -%>
```

jkroepke: > is this still correct?

No, I found this issue while I wrote the test. 😃

jkroepke:
Tests added,

ghoneycutt:
Travis failure looks like a timeout. Restarted the job.

jkroepke:
Implement changes as requested.

spec/defines/openvpn_server_spec.rb
```diff
@@ -110,6 +110,81 @@
...
+          }
+        end
+
+        it { is_expected.to contain_file('/etc/openvpn/test_server.conf').with_content(%r{^proto\s+tcp4-server$}) }
+      end
+
+      context 'when using tcp6' do
+        let(:params) do
+          {
+              'country'       => 'CO',
```
ghoneycutt: this section has four spaces instead of two and is throwing style errors at https://travis-ci.org/voxpupuli/puppet-openvpn/jobs/488140662

spec/defines/openvpn_server_spec.rb
```diff
@@ -110,6 +110,81 @@
...
+          }
+        end
+
+        it { is_expected.to contain_file('/etc/openvpn/test_server.conf').with_content(%r{^proto\s+tcp6-server$}) }
+      end
+
+      context 'when using invalid value for proto' do
+        let(:params) do
+          {
+              'country'       => 'CO',
```
ghoneycutt: same style issue here

jkroepke:
Another question. Should `tcp-server`, `tcp4-server` and `tcp6-server` also allowed? From OpenVPN side, values like `tcp*-server` are the correct values not just `tcp`.

I know that the module append `-server` automatically but it looks like a backward compatibility feature.  

From man page, only `udp[46]?` and `tcp[46]?-server` are allowed.

spec/defines/openvpn_server_spec.rb
jkroepke: Ouuch. Why rubocop is not executed here?

ghoneycutt:
I don't follow you as the `-server` part is already added as you mentioned.

jkroepke:
* From module site `proto => 'tcp-server'` is invalid while it's valid in openvpn.
* From module site `proto => 'tcp'` is valid but invalid (or deprecated) in openvpn.

I miss the consistency between openvpn and module parameters.

ghoneycutt:
We should probably not append `-server` at all, though that would need a major version bump.

jkroepke:
I would prefer to implement `-server` as valid parameter and raise a deprecation warning for the old behavior. But this should be an another PR. ok.

jkroepke:
beaker still timeouts inside travis.

ghoneycutt:
Thanks @jkroepke !

ghoneycutt:
Released in v8.1.0

PR #333, opened 2019-03-12
Adjust clients $compression type to match servers

jkroepke:
Allow newer compress option at client config since comp-lzo is deprecated

https://community.openvpn.net/openvpn/wiki/DeprecatedOptions#a--comp-lzo

### DIFF
--- a/manifests/client.pp
+++ b/manifests/client.pp
@@ -45,7 +45,7 @@
 #
 define openvpn::client (
   String $server,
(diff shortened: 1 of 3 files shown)


jkroepke:
@juniorsysadmin why it's backwards incompatible?

juniorsysadmin:
@jkroepke I did this based on the change in type without looking at the docs, have removed the label.

PR #334, opened 2019-03-15
Add tls_crypt

jkroepke:
Add `--tls-crypt` in alternative to tls auth (new feature since 2.4)


jkroepke:
Also tested on Ubuntu 18.04 LTS w/ Linux and MacOS Clients

yakatz:
Any idea when this will make it into a release? We were about to switch to this module and then found that this feature isn't available yet. Thanks.

PR #332, opened 2019-03-06
Revoke command missing on easy-rsa 3.0 #331

jasonsattler:
https://github.com/voxpupuli/puppet-openvpn/issues/331

This block of code works for me.
```
 each($revoked) |$name| {
    exec { "revoke certificate for ${name} in context of ${server}":
      command  => ". ./vars && echo yes | ./easyrsa revoke ${name} 2>&1 | grep -E 'Already revoked|was successful|not a valid certificate' && ./easyrsa gen-crl && /bin/cp -f keys/crl.pem ../crl.pem && touch revoked/${name}",
      cwd      => "/etc/openvpn/${server}/easy-rsa",
      creates  => "/etc/openvpn/${server}/easy-rsa/revoked/${name}",
      provider => 'shell',
      notify   => Service["openvpn@${server}"]
    }
  }
```

PR #338, opened 2019-05-09
Fix client cert revoke error with easyrsa 3.0

chloesoe:
In easyrsa 3.0 (used in CentOS) the command has changed. Now there is
only a single binary to run the scripts. Further the generation of CRL
also has changed; now a new crl.pem file is created in keys/crl.pem
which overrides the symlink there. So the revocation check did not work
anymore, because the crl.pem in the base directory was not checked when
a client connected.

Resolves: VSHNOPS-1537

### DIFF
--- a/manifests/revoke.pp
+++ b/manifests/revoke.pp
@@ -42,13 +42,16 @@ define openvpn::revoke (
         provider => 'shell',
       }
     }
-    default: {
+    '2.0': {
       exec { "revoke certificate for ${name} in context of ${server}":
         command  => ". ./vars && ./revoke-full ${name}; echo \"exit $?\" | grep -qE '(error 23|exit (0|2))' && touch revoked/${name}",
         cwd      => "${etc_directory}/openvpn/${server}/easy-rsa",
         creates  => "${etc_directory}/openvpn/${server}/easy-rsa/revoked/${name}",
         provider => 'shell',
       }
+    }
+    default: {
+      fail("unexepected value for EasyRSA version, got '${openvpn::easyrsa_version}', expect 2.0 or 3.0.")
     }
   }
 }
--- a/manifests/revoke.pp
+++ b/manifests/revoke.pp
@@ -25,10 +25,30 @@ define openvpn::revoke (
 
   $etc_directory = $openvpn::etc_directory
 
-  exec { "revoke certificate for ${name} in context of ${server}":
-    command  => ". ./vars && ./revoke-full ${name}; echo \"exit $?\" | grep -qE '(error 23|exit (0|2))' && touch revoked/${name}",
-    cwd      => "${etc_directory}/openvpn/${server}/easy-rsa",
-    creates  => "${etc_directory}/openvpn/${server}/easy-rsa/revoked/${name}",
-    provider => 'shell',
+  case $openvpn::easyrsa_version {
+    '3.0': {
+      exec { "revoke certificate for ${name} in context of ${server}":
+        command  => ". ./vars && ./easyrsa --batch revoke ${name}; echo \"exit $?\" | grep -qE '(error 23|exit (0|2|))' && touch revoked/${name}",
+        cwd      => "${etc_directory}/openvpn/${server}/easy-rsa",
+        creates  => "${etc_directory}/openvpn/${server}/easy-rsa/revoked/${name}",
+        provider => 'shell',
+      }
+      # `easyrsa gen-crl` does not work, since it will create the crl.pem
+      # to keys/crl.pem which is a symlinked to crl.pem in the servers etc
+      # directory
+      exec { "renew crl.pem for ${name}":
+        command  => ". ./vars && EASYRSA_REQ_CN='' EASYRSA_REQ_OU='' openssl ca -gencrl -out ../crl.pem -config ./openssl.cnf",
+        cwd      => "${openvpn::etc_directory}/openvpn/${server}/easy-rsa",
+        provider => 'shell',
+      }
+    }
+    default: {
+      exec { "revoke certificate for ${name} in context of ${server}":
+        command  => ". ./vars && ./revoke-full ${name}; echo \"exit $?\" | grep -qE '(error 23|exit (0|2))' && touch revoked/${name}",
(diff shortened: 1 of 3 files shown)


Dan33l:
Hi @chloesoe can you patch acceptance tests to tackle this issue and ensure that revoke process works ?

manifests/revoke.pp
```diff
@@ -25,10 +25,30 @@
...
+      # `easyrsa gen-crl` does not work, since it will create the crl.pem
+      # to keys/crl.pem which is a symlinked to crl.pem in the servers etc
+      # directory
+      exec { "renew crl.pem for ${name}":
+        command  => ". ./vars && EASYRSA_REQ_CN='' EASYRSA_REQ_OU='' openssl ca -gencrl -out ../crl.pem -config ./openssl.cnf",
+        cwd      => "${openvpn::etc_directory}/openvpn/${server}/easy-rsa",
+        provider => 'shell',
+      }
+    }
+    default: {
```
Dan33l: Please use a version number instead of default with easyrsa version 2.
The default can catch strange values.
Like here https://github.com/voxpupuli/puppet-openvpn/blob/master/manifests/ca.pp#L76

Dan33l:
hum ... it looks that we have an other PR about same topic #332

bastelfreak:
@chloesoe can you take a look at the failing travis jobs?

chloesoe:
> Hi @chloesoe can you patch acceptance tests to tackle this issue and ensure that revoke process works ?

@Dan33l: As I'm new in Puppet I was not able to write working acceptance tests in a reasonable time. I pushed my work I did so far. Perhaps I will have couple of hours in a month to try to get it running.
With commit b848cb91d4cc65fed215a84ef6f75d75692da084 I got a step further so a revoke test could  possibly work, because so we don't get a deprecation warning and tests start (but failing yet).

andrekeller:
@chloesoe a3c545870183481dad52a7f83a7bd61fcdf76756 should not be fixed in this PR, the proposed solution has the same issues as #312 (making key files world readable which was not the case before). PR should not address multiple issues.

chloesoe:
@andrekeller That sounds right, but acceptance tests for revoke are failing immediately without this change. And this PR only will be approved if there are acceptance tests. So should I create a separate PR for a3c5458, wait for merging, and then pull and rebase this PR so this change is applied?

andrekeller:
> @andrekeller That sounds right, but acceptance tests for revoke are failing immediately without this change. And this PR only will be approved if there are acceptance tests. So should I create a separate PR for [a3c5458](https://github.com/voxpupuli/puppet-openvpn/commit/a3c545870183481dad52a7f83a7bd61fcdf76756), wait for merging, and then pull and rebase this PR so this change is applied?

I'm not a maintainer, but I would drop this commit from this PR and wait for #312 to be merged into master. If you feel you can help with #312 even better. But it is definitely not in scope of this PR.

spec/acceptance/openvpn_spec.rb
```diff
@@ -56,7 +57,18 @@
           remote_host => $facts['networking']['ip'],
           tls_cipher  => 'TLS-DHE-RSA-WITH-AES-128-GCM-SHA256:TLS-DHE-RSA-WITH-AES-128-CBC-SHA',
         }
-      )
+
+  openvpn::client { 'vpnclientb' :
```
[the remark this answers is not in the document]
andrekeller: The permission change comes from here: https://github.com/voxpupuli/puppet-openvpn/blob/master/manifests/ca.pp#L75 
I would argue this is wrong in the first place. Why does this need to be executable? I'd propose to fix this to be root / 640.

You would then need to ensure that the revoked file is created with the correct permission, you could amend the touch (https://github.com/voxpupuli/puppet-openvpn/blob/master/manifests/revoke.pp#L29) with a chmod to ensure the proper permissions. (You could also replace this with install ```install -b -m 640 /dev/null revoked/${name}``` however I'm not sure how portable the install command is.

manifests/revoke.pp
```diff
@@ -25,10 +25,33 @@
...
+      exec { "revoke certificate for ${name} in context of ${server}":
+        command  => ". ./vars && ./easyrsa --batch revoke ${name}; echo \"exit $?\" | grep -qE '(error 23|exit (0|2|))' && touch revoked/${name}",
+        cwd      => "${etc_directory}/openvpn/${server}/easy-rsa",
+        creates  => "${etc_directory}/openvpn/${server}/easy-rsa/revoked/${name}",
+        provider => 'shell',
+      }
+      # `easyrsa gen-crl` does not work, since it will create the crl.pem
+      # to keys/crl.pem which is a symlinked to crl.pem in the servers etc
+      # directory
+      exec { "renew crl.pem for ${name}":
```
erik-frontify: This will renew the CRL on every puppet run if there is a "revoke". Can be fixed like that:

```
@@ -32,14 +32,16 @@ define openvpn::revoke (
         cwd      => "${etc_directory}/openvpn/${server}/easy-rsa",
         creates  => "${etc_directory}/openvpn/${server}/easy-rsa/revoked/${name}",
         provider => 'shell',
+        notify  => Exec["renew crl.pem for ${name}",],
       }
       # `easyrsa gen-crl` does not work, since it will create the crl.pem
       # to keys/crl.pem which is a symlinked to crl.pem in the servers etc
       # directory
       exec { "renew crl.pem for ${name}":
-        command  => ". ./vars && EASYRSA_REQ_CN='' EASYRSA_REQ_OU='' openssl ca -gencrl -out ../crl.pem -config ./openssl.cnf",
-        cwd      => "${openvpn::etc_directory}/openvpn/${server}/easy-rsa",
-        provider => 'shell',
+        command     => ". ./vars && EASYRSA_REQ_CN='' EASYRSA_REQ_OU='' openssl ca -gencrl -out ../crl.pem -config ./openssl.cnf",
+        cwd         => "${openvpn::etc_directory}/openvpn/${server}/easy-rsa",
+        provider    => 'shell',
+        refreshonly => true,
       }
     }
     '2.0': {
```

Rubueno:
@bastelfreak This can be closed.

PR #340, opened 2019-06-03
Allow puppetlabs/concat 6.x, puppetlabs/stdlib 6.x

dhoppe:


dhoppe:
> Can you avoid to switch the order of module ? Or any reason for this switch ?
I like to sort modules in alphabetically order for a better overview. Some modules have a huge list of dependencies.

Dan33l:
Ah yes ! Good reason for reordering.

What requires to increase the minimal version for concat ?

dhoppe:
@Dan33l Not really. I just used one of @alexjfisher pull requests as boilerplate to introduce some kind of standard for lower boundaries. This version of puppetlabs/concat for example introduced support for Debian 9.

Dan33l:
@dhoppe i don't know if we have a rule about minimal value requirement.
I my opinion, the lower limit should be moved only when it is required because something does not work.

Since we have acceptance tests on this module, i would prefer to keep unchanged the current minimal version. But i am also fine enough to bump the minimal version.

dhoppe:
@bastelfreak What is your opinion on this topic? I changed the lower boundaries for a lot of modules:
* https://github.com/pulls?utf8=%E2%9C%93&q=is%3Aopen+is%3Apr+user%3Avoxpupuli+archived%3Afalse+sort%3Acreated-asc+author%3Adhoppe+

bastelfreak:
I dont care about the ordering, I just try to keep the diff small and I am too lazy reorder dependencies.

----

the minimal version should only be bumped if this there is a reason for it (puppet5/6 support, debian 9 support....). But in that case we need to update the PR title and mark it as backwards-incompatible.

alexjfisher:
I'm not in favour of making major releases over something so trivial.  We've already dropped puppet 4 support (with a major version bump) and concat doesn't support puppet 5 until version 4.1.0.  As such, you couldn't consider bumping of the minimum concat version as a bugfix/cosmetic change.

dhoppe:
@alexjfisher I agree, but some told me that I need to label the pull requests as backwards-incompatible. 

I have made ~20 pull requests of this kind and it is no fun to change them every time someone makes a statement. :(

bastelfreak:
I agree with @alexjfisher here. I think the `bug` label makes more sense here instead of `backwards-incompatible`. Anybody has a different opinion?

Dan33l:
My primary trouble is we mix two changes:
 * a maintenance update
 * a bugfix.

But finally, i am fine enough if we change the title of PR and add `bug` label.

PR #344, opened 2019-07-16
Add option to disable ordering dependencies on Openvpn::Client

Bluewind:
Signed-off-by: Florian Pritz <florian.pritz@rise-world.com>

### DIFF
--- a/manifests/client_specific_config.pp
+++ b/manifests/client_specific_config.pp
@@ -12,6 +12,7 @@
 # @param dhcp_options DHCP options to push to the client.
 # @param redirect_gateway Redirect all traffic to gateway
 # @param ensure Sets the client specific configuration file status (present or absent)
+# @param manage_client_configs Manage dependencies on Openvpn::Client ressources
 #
 # @example
 #   openvpn::client_specific_config {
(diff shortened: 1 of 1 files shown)


Dan33l:
Hi @Bluewind . Can you explaine a use case ?

Bluewind:
We have a setup where we manage client certificates outside of puppet and we'd like to migrate our servers to this module. We are perfectly fine with managing the certificates with our existing workflow. Since we do not need it, we do not configure a `Client[$name]` resource for the matching  `Client_specific_config[$name]`, but currently this code results in an error because it expects such a resource to be defined. Does that explanation help?

manifests/client_specific_config.pp
```diff
@@ -12,6 +12,7 @@
 # @param dhcp_options DHCP options to push to the client.
 # @param redirect_gateway Redirect all traffic to gateway
 # @param ensure Sets the client specific configuration file status (present or absent)
+# @param manage_client_configs Manage dependencies on Openvpn::Client ressources
```
Dan33l: The description is Puppet focused.
A description more openvpn focused looks clearer IMO.

Bluewind: I've written it like this because it doesn't really do anything apart from adding the dependency. Having said that, I just looked at the client resource and I saw that it also defines a similar ordering constraint: https://github.com/voxpupuli/puppet-openvpn/blob/master/manifests/client.pp#L89

When I look at the client_specific_config resource, I see that it only adds one file. The directory of that file is created by openvpn::server. Shouldn't it be enough to order `server -> client_specific_config` then? I don't see why the client resources should be ordered before client_specific_config.

If this sounds right to you, I'd simplify my patch by removing the boolean that I introduced and then removing the ordering constraint `server->client` from client_specific_config altogether.

Dan33l: Since we do not have acceptance tests that covert this, i feel more confident to keep your patch like this.

Or, we have to add acceptance tests that covert modification of client_specific_config.

Up to you.

Bluewind: Keeping it as is is fine with me.

Dan33l: @Bluewind thank you for the PR.

PR #345, opened 2019-07-18
release 8.2.0

Dan33l:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,33 @@ All notable changes to this project will be documented in this file.
 Each new release typically also includes the latest modulesync defaults.
 These should not affect the functionality of the module.
 
+## [v8.2.0](https://github.com/voxpupuli/puppet-openvpn/tree/v8.2.0) (2019-07-19)
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -113,10 +141,6 @@ These should not affect the functionality of the module.
 
 [Full Changelog](https://github.com/voxpupuli/puppet-openvpn/compare/v7.0.0...v7.1.0)
 
-**Implemented enhancements:**
-
-- add openvpn::deploy::\(export/client\) [\#261](https://github.com/voxpupuli/puppet-openvpn/pull/261) ([to-kn](https://github.com/to-kn))
```
alexjfisher: Can you explain why this moved?

Dan33l: I did not found why.

alexjfisher: It'd be nice to find out exactly which commit broke it and why and open an issue.  Meanwhile... do you just want to use a version that works or manually fix up this PR?

Dan33l: The moved item is not moved all the times.
Without changing the version of GCG, running several times the rake task `bundle exec rake changelog` does not gives always same result. Some times the order is not the good one.

PR #350, opened 2019-08-19
Fixing issue #211; allow to not manage client service or correctly manage it

### LINKED ISSUE #211: does not work with systemd and severals VPN created
Tried running it on Centos 7 latest and it doesn't start the openvpn process because of the new openvpn systemd logic

openvpn@service where service=service.conf inside /etc/openvpn.

erik-frontify:
Fixes #211 

Should manage the service for openvpn client service correctly. Service restart on configuration changes does _not_ work due to dependency cycles I wasn't able to solve.

Only tested on Amazon linux (like RedHat 7).


bastelfreak:
@erik-frontify can you please have a look at the failing travis jobs?

erik-frontify:
@bastelfreak, OK found the issue with the quoted variables and how to solve that. Fixed follow up issues as well.
If you see a way to reload configuration on configuration changes without running into dependency cycles that would be great. I tried to solve it but failed...

Valantin:
Since deploy resorce was deprecated this was obsolete.

PR #355, opened 2019-09-26
easy-rsa 3.0 by default for FreeBSD

olevole:
easy-rsa2 was removed from FreeBSD ports tree, see:
https://svnweb.freebsd.org/ports?view=revision&revision=r504939

switch to easy-rsa 3, tested on FreeBSD 12.0-RELEASE, FreeBSD 13-CURENT


bastelfreak:
Thansk for the PR @olevole. Could you add the supported FreeBSD versions into the metadata.json?

bastelfreak:
@olevole can you take a look at the failing spec tests?

olevole:
@bastelfreak Yeah, I have already seen. It takes some work, I'm in the process

PR #356, opened 2019-10-19
Clean up acceptance spec helper

ekohl:


ekohl:
Rebased. The Debian 9 build failed with a timeout to the forge so I restarted that. The others were green so I'd expect that to pass as well.

Issue #263, opened 2018-01-31
Predictable Interface Names on Debian 9 (stretch) not working


duckpuppy:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 5.3.3
- Ruby: 2.4.2
- Distribution: Debian 9 (stretch)
- Module version: 7.1.0

## How to reproduce (e.g Puppet code you use)

Simply include the module and declare a server

## What are you seeing

`Error: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Unknown variable: '::ipaddress_eth0'. at /etc/puppetlabs/code/environments/production/modules/openvpn/manifests/server.pp:451:47  at /etc/puppetlabs/code/environments/production/manif
ests/openvpn.pp:10 on node dagobah.pondnet`

## What behaviour did you expect instead
No error

## Any additional information you'd like to impart

As of Debian 9, predictable interface names are used by default.  That means that there's no `eth0`, etc.  Instead, the interfaces are named by BIOS information.  My primary interface is named `eno1`.  In fact... facter reports an empty `interfaces` fact. There's no fact named `::ipaddress_eth0`, but there is one named `ipaddress`.

duckpuppy:
And after doing what I thought was a decent search of things, I found the `local` configuration option.

Rathios:
This default is also an issue on RedHat/CentOS 7, which uses a different interface naming scheme.

From `openvpn::server`:
```
 String $local = $facts['ipaddress_eth0'],
```

Which results in
```
Error while evaluating a Resource Statement, Openvpn::Server[servername]: parameter 'local' expects a String value, got Undef
```

A more sane default would be just `$facts['ipaddress']`

PR #367, opened 2019-12-06
modulesync 2.10.1

dhoppe:


dhoppe:
Ubuntu uses a different name for the interface: 
* https://github.com/camptocamp/facterdb/blob/master/facts/3.11/ubuntu-18.04-x86_64.facts#L255

@bastelfreak Do we want to use ```$facts['networking']['ip']``` to fix the CI pipeline? This would also fix #263 and the user is still able to override the value of ```$local``` at ```manifests/server.pp```.

manifests/server.pp
```diff
@@ -153,7 +153,7 @@
   Optional[String] $group                                           = undef,
   Boolean $ipp                                                      = false,
   Boolean $duplicate_cn                                             = false,
-  String $local                                                     = $facts['ipaddress_eth0'],
+  String $local                                                     = $facts['networking']['interfaces']['eth0']['ip'],
```
bastelfreak: it probably makes more sense to use the default ip address
```suggestion
  String $local                                                     = $facts['networking']['ip'],
```

dhoppe:
The tests are still failing, because Facer 3.14 provides the wrong hostname, domain and fqdn for Arch Linux. I have fixed this via https://github.com/camptocamp/facterdb/pull/146.

PR #369, opened 2019-12-13
Enable revocation when easyrsa version 3.0 is used

### LINKED ISSUE #331: Revoke command missing on easy-rsa 3.0

- Puppet: 5.5.10
- Ruby: ruby 2.4.5p335 (2018-10-18 revision 65137) [x86_64-linux]
- Distribution: centos
- Module version: any
- easy-rsa version: 3.0.3-1.el7.noarch

## What are you seeing
easy-rsa change the way they revoke.  
They no longer have `./revoke-full`

It is now `./easyrsa revoke`
When I do it manually it will ask you to confirm that you want to delete by typing `yes`.

Then I had to create a new `/etc/openvpn/<server>/crl.pem` by running `./easyrsa gen-crl`
Then restart the openvpn server

Rubueno:
This change makes it possible to revoke certificates when easyrsa version 3.x is used. There currently is an open PR open for this issue, however this has not had an update since March 2019, therefore I am opening this PR. See: #338 

The `remote_host` parameter is set, because the expected value of
`foo.example.com` is not evaluated correctly on ArchLinux.

Fixes #331

### DIFF
--- a/manifests/revoke.pp
+++ b/manifests/revoke.pp
@@ -25,10 +25,29 @@ define openvpn::revoke (
 
   $etc_directory = $openvpn::etc_directory
 
+  $revocation_command = $openvpn::easyrsa_version ? {
+    '2.0' => ". ./vars && ./revoke-full ${name}; echo \"exit $?\" | grep -qE '(error 23|exit (0|2))' && touch revoked/${name}",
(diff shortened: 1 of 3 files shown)


bastelfreak:
thanks for the PR!

PR #372, opened 2019-12-24
name of openssl.cnf changed in easy-rsa 3 or debian buster

whohoho:


ghoneycutt:
This might work for Debian Buster, though would not work for other releases. Suggest using a conditional to set a variable for the file name based on the platform.

NITEMAN:
For buster to work the only needed change seems to setting easy-rsa to the correct version:
```yaml
openvpn::default_easyrsa_ver: '3.0'
```

PR #379, opened 2020-05-04
Add buster support

NITEMAN:
#### Pull Request (PR) description
Add Debian 10 (buster) support

#### This Pull Request (PR) fixes the following issues
Fixes #372 (superseeds)

PR #380, opened 2020-05-05
modulesync 2.12.0

bastelfreak:
modulesync 2.12.0

### DIFF
--- a/spec/spec_helper_acceptance.rb
+++ b/spec/spec_helper_acceptance.rb
@@ -1,35 +1,38 @@
 require 'voxpupuli/acceptance/spec_helper_acceptance'
 
 configure_beaker do |host|
-  # Configure all nodes in nodeset
-  install_server_packages = %(
+  manifest = <<-PUPPET
     if $facts['os']['name'] == 'CentOS' {
       package { 'epel-release':
         ensure => present,
       }
     }
 
-    $package_name = $facts['os']['family'] ? {
+    $netcat_package_name = $facts['os']['family'] ? {
       'Debian' => 'netcat-openbsd',
       'RedHat' => 'nc',
       default  => 'netcat',
     }
-    package { $package_name:
-      ensure => present,
-    }
-  )
-  apply_manifest_on(hosts_as('vpnserver'), install_server_packages, catch_failures: true)
 
-  install_client_packages = %(
-    if $facts['os']['name'] == 'CentOS' {
-      package { 'epel-release':
+    node /^vpnserver\./ {
+      package { $package_name:
         ensure => present,
       }
     }
 
-    package { ['tar','openvpn'] :
-      ensure => present,
+    node /^vpnclienta\./ {
+      package { ['tar','openvpn'] :
+        ensure => present,
+      }
     }
-  )
-  apply_manifest_on(hosts_as('vpnclienta'), install_client_packages, catch_failures: true)
+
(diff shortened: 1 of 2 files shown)


spec/spec_helper_acceptance.rb
```diff
@@ -1,45 +1,35 @@
...
+    $package_name = $facts['os']['family'] ? {
+      'Debian' => 'netcat-openbsd',
+      'RedHat' => 'nc',
+      default  => 'netcat',
+    }
+    package { $package_name:
+      ensure => present,
+    }
+  )
+  apply_manifest_on(hosts_as('vpnserver'), install_server_packages, catch_failures: true)
```
ekohl, line 21: This is not correct. It runs it for every host, rather than just once on a selected set of nodes. This is because `configure_beaker` iterates over every host. I haven't thought about a clean way to do this.

spec/spec_helper_acceptance.rb
```diff
@@ -1,45 +1,35 @@
-require 'beaker-rspec'
-require 'beaker-puppet'
-require 'beaker/puppet_install_helper'
-require 'beaker/module_install_helper'
+require 'voxpupuli/acceptance/spec_helper'
```
ekohl, line 1:
```suggestion
require 'voxpupuli/acceptance/spec_helper_acceptance'
```

Now that I see it, perhaps it was redundant to add `_acceptance` in the filename given the directory name.

bastelfreak: true, but that keeps it in sync with the spec_helper_acceptance.rb that we've in our modules. So maybe not too bad.

ekohl: I'll keep this in mind for a 1.0 release.

ekohl:
I pushed a commit that I haven't tested, but I hope is a good direction for multi node setups.

ekohl:
It fails on CentOS 6 because it looks like the hostname can't be set there, so it's `localhost` instead of `vpnserver`. I'm not sure how to proceed with this. Maybe we can read the roles from the `host` object and apply it based on that?

NITEMAN:
Awaiting this to rebase #379 and sending other PR, may I help somehow?

ekohl:
The alternative is to just install all packages on every node rather than some on the server and some on the client. I don't think that's an issue.

spec/spec_helper_acceptance.rb
```diff
@@ -1,45 +1,38 @@
...
-          ensure => present,
-        }
+    node /^vpnclienta\./ {
+      package { ['tar','openvpn'] :
+        ensure => present,
       }
+    }
 
-      package { ['tar','openvpn'] :
+    # CentOS 6 in docker doesn't get a hostname - install all packages
```
ekohl, line 29: This is the workaround I decided to go with

spec/spec_helper_acceptance.rb
```diff
@@ -1,45 +1,38 @@
...
-        default  => 'netcat',
-      }
+    $netcat_package_name = $facts['os']['family'] ? {
+      'Debian' => 'netcat-openbsd',
+      'RedHat' => 'nc',
+      default  => 'netcat',
+    }
+
+    node /^vpnserver\./ {
       package { $package_name:
```
NITEMAN, line 18: This is a leftover, fixed in #379

PR #385, opened 2020-10-15
Update code to set status parameter optional

smutel:

Currently the openvpn option called status is mandatory and set with /var/log/openvpn/${name}-status.log as default.
However this parameter is not mandatory in the openvpn config. If this parameter is omitted, every log will be append to openvpn.log.

### DIFF
--- a/spec/defines/openvpn_server_spec.rb
+++ b/spec/defines/openvpn_server_spec.rb
@@ -709,6 +709,21 @@ describe 'openvpn::server' do
           it { is_expected.to contain_file("#{server_directory}/test_server.conf").with_content(%r{^client\-disconnect\s+"#{server_directory}/test_server/scripts/disconnect\.sh"$}) }
         end
 
+        context 'when not using status log' do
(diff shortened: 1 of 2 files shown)


kenyon (approved):

kenyon:
Would be nice to also have a test for this.

smutel:
I am not an expect with puppet and/or ruby but I can try ...
Could you please guide me ?

kenyon:
> I am not an expect with puppet and/or ruby but I can try ...
> Could you please guide me ?

In this file https://github.com/voxpupuli/puppet-openvpn/blob/master/spec/defines/openvpn_server_spec.rb  you want to have a test that verifies that the `status` option does not appear in the config file when `status_log` is set to an empty string.

smutel:
I added one test. Tell me if it's ok.

kenyon (approved):

smutel:
Did you plan to create a new tag soon ?

PR #392, opened 2020-12-06
fix(client): Handle expire value for easyrsa version 3

Turgon37:
EasyRSA version 3 and above has renamed some environment vars and then source by default the "vars" file. So client.pp allow override the "expire" for certificates but do not handle this change.

You can notice here that the "vars" template handle correctly the change https://github.com/voxpupuli/puppet-openvpn/blob/master/templates/vars-30.epp#L59 but not client.pp

### DIFF
--- a/manifests/client.pp
+++ b/manifests/client.pp
@@ -107,6 +107,9 @@ define openvpn::client (
         '3.0': {
           $env_expire = "EASYRSA_CERT_EXPIRE=${expire} EASYRSA_NO_VARS=1"
         }
+        default: {
+          fail("unexepected value for EasyRSA version, got '${openvpn::easyrsa_version}', expect 2.0 or 3.0.")
+        }
       }
     } else {
       warning("Custom expiry time ignored: only integer is accepted but ${expire} is given.")
--- a/manifests/client.pp
+++ b/manifests/client.pp
@@ -100,7 +100,14 @@ define openvpn::client (
 
   if $expire {
     if is_integer($expire) {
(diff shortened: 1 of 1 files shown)


manifests/client.pp
```diff
@@ -100,7 +100,14 @@
 
   if $expire {
     if is_integer($expire) {
-      $env_expire = "KEY_EXPIRE=${expire}"
+      case $openvpn::easyrsa_version {
+        '2.0': {
+          $env_expire = "KEY_EXPIRE=${expire}"
+        }
+        '3.0': {
+          $env_expire = "EASYRSA_CERT_EXPIRE=${expire} EASYRSA_NO_VARS=1"
```
Turgon37, line 108: I've used EASYRSA_NO_VARS because by default easy-rsa source the "vars" file at startup and then overwrite variable given by shell environment. With EASYRSA_NO_VARS, easy-rsa do not source anything, but because we use ". vars" in exec resource, all vars are already provisionned in shell environment

 https://github.com/OpenVPN/easy-rsa/blob/master/easyrsa3/easyrsa#L2481

bastelfreak:
Hi, thanks for the PR! Please rebase after https://github.com/voxpupuli/puppet-openvpn/pull/393 got merged

Turgon37:
> Hi, thanks for the PR! Please rebase after #393 got merged

Hello It's done :)

kenyon (approved):

Turgon37:
Hello, Is this ok to be merged ?

PR #396, opened 2021-05-11
add service-notify to scheduled crl_auto_renew exec

zivis:
If using crl_auto_renew and the crl is recreated, the service has to be reloaded, 
otherwise clients can't connect if crl is getting verified. (seen on centos7)

no issue created, but i can if it helps


root-expert:
Hello! Thanks for the PR.

According to the [documentation](https://openvpn.net/community-resources/revoking-certificates/) when crl-verify is set, the crl file will be re-read every time a new client tries to connect.

So you can update the crl file without having to reload the deamon.
Maybe the issue is somewhere else?

Valantin:
If needed reopen and rebase

PR #410, opened 2021-09-16
Use Deferred functions instead facts

### LINKED ISSUE #409: Warning: Fact value '...' with the value length: '5274' exceeds the value length limit: 4096

- Puppet: 7.11.0
- Ruby: 2.7.3p183
- Distribution: Ubuntu 20.04
- Module version: 9.0.0

## What are you seeing

> $ puppet agent -t
> Warning: Fact value '...' with the value length: '5274' exceeds the value length limit: 4096

... contains a while certificate

No warnings

## Output log

> $ puppet agent -t
> Warning: Fact value '...' with the value length: '5274' exceeds the value length limit: 4096

## Any additional information you'd like to impart
The warning introduced in puppet 7.10 

https://puppet.com/docs/puppet/7/release_notes_puppet.html#enhancements_puppet_x-7-10-0-PUP-11088

Looking at https://github.com/voxpupuli/puppet-openvpn/issues/322 , I would recommend to introduce a breaking change by removing the facts (and deploy logic) and bump the major version.

jkroepke:
Replace facts with deferred functions

Fixes #409


jkroepke:
I tried to follow https://github.com/voxpupuli/puppet-gitlab_ci_runner/blob/11c2d0ad3f96f0f5146e82672bebb35c416eb183/spec/classes/gitlab_ci_runner_spec.rb#L10-L17 but I don't know why tests not working here.

root-expert:
Hello @jkroepke take a look at the failing rubocop offences. Fixing them should allow the unit tests to run

jkroepke:
@root-expert I agree, the error recently accours since the latest push

root-expert:
I believe this has to do with rspec scoping. Try moving your `before` to the `context` on line 66. That may solve it.

lib/puppet/functions/openvpn/file_content.rb
```diff
@@ -0,0 +1,20 @@
+# A function that return a file content at apply time
+Puppet::Functions.create_function(:'openvpn::file_content') do
+  # @param filename Path of file to cat
+  # @return [String] Returns the file content
+
```
alexjfisher, line 5: Does this empty line break puppet-strings?

jkroepke: I hope a test covers this.

jkroepke:
```suggestion

```

jkroepke:
> I believe this has to do with rspec scoping. Try moving your `before` to the `context` on line 66. That may solve it.

I try that before, too. But on the gitlab-ci module, its on top scope, too.

lib/puppet/functions/openvpn/file_content.rb
alexjfisher: Puppet-lint [covers this](https://github.com/rodjek/puppet-lint/pull/931) for puppet manifests, but I'm not aware of anything that would have warned you here, (and I'm only guessing it would break).

spec/defines/openvpn_deploy_export_spec.rb
```diff
@@ -18,16 +29,41 @@
...
+        facts
       end
       let(:title) { 'test_client' }
       let(:params) { { 'server' => 'test_server' } }
 
       it { is_expected.to compile.with_all_deps }
 
       context 'exported resources' do
+        before do
+          allow(File).to receive(:exist?).and_call_original
```
jkroepke, line 41: Any idea, why this wont work?

jkroepke:
Initially, there was a typo inside the server_path switch case.

On top, I added the export resource to acceptance tests.

root-expert (approved):

root-expert:
Do you think it makes sense to switch to the function only for the key and keep the facts for the rest?

jkroepke:
> Do you think it makes sense to switch to the function only for the key and keep the facts for the rest?

That would not fix https://github.com/voxpupuli/puppet-openvpn/issues/409 . I guess the default value of 4096 for fact limit size has some valid reason.

> People may still need to use the facts (not the key one since it's a security risk)

I dont know a valid case of this kind of facts. Inside puppet, they could use the new deferred function.

In case, deferred function does not fit because they re-use the facts outside of puppet, I should recommend that the peoples creates the fact on they own

OR

create a dedicated puppet module like voxpupuli/openvpn-facts to make the facts optional. (including documentation that it need adjustments on puppet default settings).

bastelfreak (approved):

jkroepke:
Hi,

are 2 approvals fine to merge this?

lib/puppet/functions/openvpn/file_content.rb
```diff
@@ -0,0 +1,19 @@
+# A function that return a file content at apply time
```
ghoneycutt, line 1: This function should have tests.

ghoneycutt (changes requested):
Please add tests for the new function. Stdlib is a great place to look for those types of tests.

lib/puppet/functions/openvpn/file_content.rb
jkroepke: @ghoneycutt Tests added.

bastelfreak (approved):

jameskirsop:
@jkroepke, did you test this with the `openvpn::deploy::client` class? I'm trying to get to the bottom of #421.

jkroepke:
Long time ago, but I through i test it. 

There is also an acceptance test inside the CI, that test the mention class. https://github.com/voxpupuli/puppet-openvpn/pull/410/files#diff-79341f13d5586cff43fd13a0132d5918a8b9a6c3b36cbcdd52b617a598a9d508R134

PR #412, opened 2021-09-18
Disable compression and set cipher to AES-256-GCM by default

jkroepke:
Since https://github.com/voxpupuli/puppet-openvpn/pull/410 would introduce a breaking change, I would recommend to include this breaking change, too.

OpenVPN is going to remove the compression feature. I would disable compression by default. That results that new OpenVPN servers are future proof by default.

OpenVPN not longer recommends to define a cipher, since OpenVPN use (and enforce in upcoming version) cipher negotiation. For 2.4 compability, the 'AES-256-CBC' isn't a good default anymore. Using 'AES-256-GCM' as default is much better here, since it's the default cipher of the cipher negotiation.

> Sep 16 22:52:56 PI openvpn[961]: DEPRECATED OPTION: --cipher set to 'AES-256-CBC' but missing in --data-ciphers (AES-256-GCM:AES-128-GCM). Future OpenVPN version will ignore --cipher for cipher negotiations. Add 'AES-256-CBC' to --data-ciphers or change --cipher 'AES-256-CBC' to --data-ciphers-fallback 'AES-256-CBC' to silence this warning.

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -7,10 +7,10 @@
 # @param organization Organization to be used for the SSL certificate, mandatory for server mode.
 # @param email Email address to be used for the SSL certificate, mandatory for server mode.
 # @param remote List of OpenVPN endpoints to connect to.
-# @param remote_random_hostname OpenVPN will prepend a random string (6 bytes, 12 hex characters) to hostname to prevent DNS caching. For example, "foo.example.com" would be modified to "<random-chars>.foo.example.com". 
+# @param remote_random_hostname OpenVPN will prepend a random string (6 bytes, 12 hex characters) to hostname to prevent DNS caching. For example, "foo.example.com" would be modified to "<random-chars>.foo.example.com".
 # @param remote_random When multiple ${remote} address/ports are specified, initially randomize the order of the list as a kind of basic load-balancing measure.
 # @param common_name Common name to be used for the SSL certificate
-# @param compression Which compression algorithim to use
+# @param compression Which compression algorithm to use. This parameter is deprecated in OpenVPN 2.5.
 # @param dev TUN/TAP virtual network device
 # @param user Group to drop privileges to after startup
 # @param group User to drop privileges to after startup
@@ -151,7 +151,7 @@ define openvpn::server (
   Boolean $remote_random_hostname                                   = false,
   Boolean $remote_random                                            = false,
   String $common_name                                               = 'server',
-  String $compression                                               = 'comp-lzo',
+  String $compression                                               = '',
   String $dev                                                       = 'tun0',
   String $user                                                      = 'nobody',
   Optional[String] $group                                           = undef,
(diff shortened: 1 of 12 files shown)


manifests/server.pp
```diff
@@ -151,7 +151,7 @@
   Boolean $remote_random_hostname                                   = false,
   Boolean $remote_random                                            = false,
   String $common_name                                               = 'server',
-  String $compression                                               = 'comp-lzo',
+  String $compression                                               = '',
```
root-expert, line 154: You may need to update the erb/epp file to reflect this change
```suggestion
  Optional[String[1]] $compression                                  = undef,
```

manifests/server.pp
```diff
@@ -212,7 +212,7 @@
   String $key_name                                                  = '',
   String $key_ou                                                    = '',
   String $verb                                                      = '',
-  String $cipher                                                    = 'AES-256-CBC',
+  String $cipher                                                    = '',
```
root-expert, line 215: You may need to update the erb/epp file to reflect this change
```suggestion
  Optional[String[1]] $cipher                                       = undef,
```

root-expert (commented):
Can you provide some documentation from OpenVPN docs if available for these changes?

jkroepke:
> Can you provide some documentation from OpenVPN docs if available for these changes?

https://community.openvpn.net/openvpn/wiki/DeprecatedOptions#Option:--compress

https://build.openvpn.net/man/openvpn-2.5/openvpn.8.html (see --compress)

Background: https://community.openvpn.net/openvpn/wiki/VORACLE

jkroepke:
Ubuntu 16.04 is failing.

Should I fixed that by enrich the default for Ubuntu 16.04 or can we drop Ubuntu 16.04, since its [EOL](https://wiki.ubuntu.com/Releases)?

root-expert:
> Ubuntu 16.04 is failing.
> 
> Should I fixed that by enrich the default for Ubuntu 16.04 or can we drop Ubuntu 16.04, since its [EOL](https://wiki.ubuntu.com/Releases)?

We can drop it along with Debian 9. Are you able to provide another PR and rebase this one on top of it?

jkroepke:
> We can drop it along with Debian 9.

Why? https://wiki.debian.org/LTS. It is still supported until 2022

jkroepke:
I ll rebase, if the other PR is in master.

jkroepke:
@root-expert 

I figure out, we could also drop `--ns-cert-type` (default to true in this module) and set `--remote-cert-tls` (default to false in this module) to true, too.

https://community.openvpn.net/openvpn/wiki/DeprecatedOptions#Option:--ns-cert-type

Can I include this change here OR should I create a new PR?

root-expert:
We need a new PR for that @jkroepke , so the users are aware for the change through the Changelog

root-expert:
Can you rebase this on top of #413 ?

root-expert (approved):

root-expert:
This PR also contains:

* #413

PR #419, opened 2021-10-04
Allowed openvpn::ca declared in openvpn::server to set crl_days parameter

Deroin:
I noticed that `openvpn::server` may instantiate `openvpn::ca` where `crl_days` is not configurable. This will cause `crl_days` to always be 30 days which can be annoying when not revoking certificates or regenerating the CRL within those 30 days.

Thus I extended this to allow a value for `openvpn::server::crl_days` to be set which is then forwarded to `openvpn::ca`.


root-expert (changes requested):
Thank you for the PR! 

Can you update [openvpn_server_spec](https://github.com/voxpupuli/puppet-openvpn/blob/master/spec/defines/openvpn_server_spec.rb#L357) by adding the new parameter?

root-expert:
@Deroin ensure `crl_days` is rendered correctly on the file inside the specs please.

Deroin:
> @Deroin ensure `crl_days` is rendered correctly on the file inside the specs please.

I'm sorry, but I don't know what you mean by that. I basically checked where `key_expire` is mentioned and extended the spec in the same way for `crl_days`.

root-expert (approved):
My bad, seems okay to me

PR #432, opened 2022-04-20
Add support for easyrsa X509 DN mode 'cn_only'

jkroepke:

In most of the time, the `Country/Province/City/Org/OU/email` are pointless for OpenVPN servers. easy-rsa introduce cn_only ad DN mode to omit `Country/Province/City/Org/OU/email` values.

```
# Define X509 DN mode.
# This is used to adjust what elements are included in the Subject field as the DN
# (this is the "Distinguished Name.")
# Note that in cn_only mode the Organizational fields further below are not used.
#
# Choices are:
#   cn_only  - use just a CN value
#   org      - use the "traditional" Country/Province/City/Org/OU/email/CN format

#set_var EASYRSA_DN	"cn_only"
```

This PR also remove the manual generation of the in favor of `easyrsa gen-crl` command, since the old variant was not compatible cn_only mode. `gen-crl` put the crl.pem in private folder of the easyrsa PKI folder, a manual copy is needed.

This PR is in draft mode for get some early reviews. Since this PR is in conflict with #431, I propose the focus should be on #431. After merge of #431, I'll do a rebase here.

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -1,6 +1,7 @@
 #
 # @summary This define creates the openvpn server instance which can run in server or client mode.
 #
+# @param dn_mode EasyRSA X509 DN mode.
 # @param country Country to be used for the SSL certificate, mandatory for server mode.
 # @param province Province to be used for the SSL certificate, mandatory for server mode.
 # @param city City to be used for the SSL certificate, mandatory for server mode.
@@ -146,6 +147,7 @@
 #   }
 #
 define openvpn::server (
+  Enum['org','cn_only'] $dn_mode                                    = 'org',
   Optional[String[1]] $country                                      = undef,
   Optional[String[1]] $province                                     = undef,
   Optional[String[1]] $city                                         = undef,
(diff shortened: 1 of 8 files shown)


manifests/server.pp
```diff
@@ -404,11 +409,16 @@
           }
           '3.0': {
             exec { "renew crl.pem on ${name}":
-              command  => ". ./vars && EASYRSA_REQ_CN='' EASYRSA_REQ_OU='' openssl ca -gencrl -out ${server_directory}/${name}/crl.pem -config ${server_directory}/${name}/easy-rsa/openssl.cnf",
+              command  => "./easyrsa gen-crl && cp ./keys/crl.pem ${server_directory}/${server}/crl.pem",
```
bastelfreak, line 412: does this still work with easy-rsa 2 or does it already drop support for it?

jkroepke: No, this is easy-rsa 3.0 exclusive which is fine at the line, because the exec resource is only executed if easyrsa_version is 3.0. Take note of the context it's inside a switch/case.

jkroepke:
Test failing since factordb 1.16.0 - take a look here https://github.com/voxpupuli/puppet-openvpn/actions/runs/2201924327 to see results with pinned factordb

bastelfreak (approved):

Gemfile
```diff
@@ -27,6 +27,8 @@ end
 
 gem 'rake', :require => false
 gem 'facter', ENV['FACTER_GEM_VERSION'], :require => false, :groups => [:test]
+# https://github.com/voxpupuli/facterdb/pull/244#issuecomment-1105120825
+gem 'facterdb', '= 1.15.0', :require => false, :groups => [:test]
```
bastelfreak, line 31: can you please undo this? We were able to figure out a proper fix for rspec-puppet-facts and released it a few minutes ago.

bastelfreak (approved):

PR #436, opened 2022-05-07
Make ecdh-curve optional

jkroepke:

Make ecdh-curve optional, if no dh key is defined.

In OpenVPN 2.5 (I initially test the EC keys with OpenVPN 2.4), define ecdh-curve will throw an warning

> Consider setting groups/curves preference with tls-groups instead of forcing a specific curve with ecdh-curve.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -1339,11 +1339,11 @@ Default value: `'secp384r1'`
 
 ##### <a name="ecdh_curve"></a>`ecdh_curve`
 
-Data type: `String`
+Data type: `Optional[String[1]]`
 
 Define the named curve for ECDH key exchange, used if ssl_key_algo is ec, ed
 
-Default value: `'secp384r1'`
+Default value: ``undef``
 
 ##### <a name="topology"></a>`topology`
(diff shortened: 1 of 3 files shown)


bastelfreak (approved):

jkroepke:
BC change cloud be discussable, it would not break an existing OVPN setup.

jkroepke:
@bastelfreak BC change cloud be discussable, it would not break an existing OVPN setup.

jkroepke:
Can I help here to move forward here?

bastelfreak:
This will purge `ecdh-curve` from the config. doesn't that effect existing configurations? Or is secp384r1 the default value?

jkroepke:
If `ecdh-curve` is purged from config, the defaults will be inherit from `tls-groups` option. The default is `X25519:secp256r1:X448:secp521r1:secp384r1`

> If you do not force the ecdh curve by using --ecdh-curve, the groups for ecdh will also be picked from this list.

https://build.openvpn.net/man/openvpn-2.5/openvpn.8.html

PR #437, opened 2022-06-21
Fix path for crl_auto_renew with easy_rsa 3.0

cible:
Small patch to fix the path of crl.pem introduced by https://github.com/voxpupuli/puppet-openvpn/pull/432 by @jkroepke
The cp fail with : 
```
'./easyrsa gen-crl && cp ./keys/crl.pem /etc/openvpn/10.9.200.0 255.255.255.0/crl.pem' returned 1 instead of one of [0]
change from 'notrun' to ['0'] failed: './easyrsa gen-crl && cp ./keys/crl.pem /etc/openvpn/10.9.200.0 255.255.255.0/crl.pem' returned 1 instead of one of [0] (corrective)
```
Certainly a cut and paste problem. In manifests/revoke.pp, $server is the correct variable but not in server.pp

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -409,7 +409,7 @@ define openvpn::server (
           }
           '3.0': {
             exec { "renew crl.pem on ${name}":
-              command  => "./easyrsa gen-crl && cp ./keys/crl.pem ${server_directory}/${server}/crl.pem",
+              command  => "./easyrsa gen-crl && cp ./keys/crl.pem ${server_directory}/${name}/crl.pem",
               cwd      => "${server_directory}/${name}/easy-rsa",
               provider => 'shell',
               schedule => "renew crl.pem schedule on ${name}",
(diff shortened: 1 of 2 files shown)


jkroepke:
Consider to add an test case for this. :)

jkroepke:
@cible in fact you have to test if the puppet exec with command correct exists.

https://github.com/voxpupuli/puppet-openvpn/blob/fe0e82db1cf1f064393175af69f7db04ff7e728e/spec/defines/openvpn_server_spec.rb#L479

With exec resource itself is already including. The command parameter needs to be added, like:

```ruby
it { is_expected.to contain_exec('renew crl.pem on test_server').with('command' => './easyrsa gen-crl && cp ./keys/crl.pem /etc/openvpn/server/crl.pem') } } 
```

In worse, an additional condition for easy-rsa 2.0 and 3.0 is needed

jkroepke (approved):
lgtm

smortex:
Looks good! Can you please combine your 7 commits in a single one?

From your working directory:

```
git rebase -i 2b2168e796a58a851b13431e438a6fbf24b8d69f # Interactively rewrite history
```

This will bring your editor listing all your commits.  Keep the first line unchanged, and replace `pick` with `squash` for all subsequent lines.  Save and quit, your editor will pop again and allow you to combine all the commit messages in a single one.

Then push your updated branch:

```
git push -f            # Send the changes (-f is required because we re-wrote history)
```

PR #444, opened 2023-05-22
Support data-ciphers and data-ciphers-fallback options

rvm-peercode:

Support `data-ciphers` and `data-ciphers-fallback` options for OpenVPN 2.5.

#### This Pull Request (PR) fixes the following issues: n/a

### DIFF
--- a/manifests/client.pp
+++ b/manifests/client.pp
@@ -14,6 +14,8 @@
 # @param remote_host  The IP or hostname of the openvpn server service.
 # @param cipher Cipher to use for packet encryption
 # @param tls_cipher TLS Ciphers to use
+# @param data_ciphers Ciphers to allow for packet encryption
+# @param data_ciphers_fallback Cipher to use if peer cipher config cannot be determined
 # @param resolv_retry  How many seconds should the openvpn client try to resolve the server's hostname
 # @param auth_retry Controls how OpenVPN responds to username/password verification errors such as the client-side response to an AUTH_FAILED message from the server or verification failure of the private key password.
 # @param verb Level of logging verbosity
@@ -61,6 +63,8 @@ define openvpn::client (
   Boolean $pam                                         = false,
   String $cipher                                       = 'AES-256-GCM',
   String $tls_cipher                                   = 'TLS-DHE-RSA-WITH-AES-256-GCM-SHA384:TLS-DHE-RSA-WITH-AES-256-CBC-SHA256:TLS-DHE-RSA-WITH-AES-128-GCM-SHA256:TLS-DHE-RSA-WITH-AES-128-CBC-SHA256',
+  String $data_ciphers                                 = 'AES-256-GCM:AES-128-GCM',
+  Optional[String] $data_ciphers_fallback              = undef,
   Boolean $authuserpass                                = false,
   Hash $setenv                                         = {},
   Hash $setenv_safe                                    = {},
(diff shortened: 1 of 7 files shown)


manifests/client.pp
```diff
@@ -61,6 +63,8 @@
   Boolean $pam                                         = false,
   String $cipher                                       = 'AES-256-GCM',
   String $tls_cipher                                   = 'TLS-DHE-RSA-WITH-AES-256-GCM-SHA384:TLS-DHE-RSA-WITH-AES-256-CBC-SHA256:TLS-DHE-RSA-WITH-AES-128-GCM-SHA256:TLS-DHE-RSA-WITH-AES-128-CBC-SHA256',
+  Optional[String] $data_ciphers                       = 'AES-256-GCM:AES-128-GCM',
```
smortex, line 66: `Optional` with a default value cannot be forced to undef. Since we test the value against `''`, it make sense to only allow `String`:

```suggestion
  String $data_ciphers                                 = 'AES-256-GCM:AES-128-GCM',
```

smortex (commented):
Some minor data types issues.  I added in-line comments to address them.

I would suggest to stick to String in this PR to match the rest of the module for now, and if you are up to it work on improving the module so that OpenVPN use the default values (i.e. no parameter is output) when no explicit value is provided instead of when a `''` is passed in another PR that will go in a new major release.

manifests/client.pp
```diff
@@ -61,6 +63,8 @@
   Boolean $pam                                         = false,
   String $cipher                                       = 'AES-256-GCM',
   String $tls_cipher                                   = 'TLS-DHE-RSA-WITH-AES-256-GCM-SHA384:TLS-DHE-RSA-WITH-AES-256-CBC-SHA256:TLS-DHE-RSA-WITH-AES-128-GCM-SHA256:TLS-DHE-RSA-WITH-AES-128-CBC-SHA256',
+  String $data_ciphers                                 = 'AES-256-GCM:AES-128-GCM',
```
yakatz, line 66: This is potentially a breaking change. Can it default to the same value as `ciphers` instead of a hard coded value that could be different than what the user expects?

manifests/client.pp
```diff
@@ -61,6 +63,8 @@
   Boolean $pam                                         = false,
   String $cipher                                       = 'AES-256-GCM',
   String $tls_cipher                                   = 'TLS-DHE-RSA-WITH-AES-256-GCM-SHA384:TLS-DHE-RSA-WITH-AES-256-CBC-SHA256:TLS-DHE-RSA-WITH-AES-128-GCM-SHA256:TLS-DHE-RSA-WITH-AES-128-CBC-SHA256',
+  Optional[String] $data_ciphers                       = 'AES-256-GCM:AES-128-GCM',
+  Optional[String] $data_ciphers_fallback              = undef,
```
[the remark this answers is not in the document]
yakatz: I would also like to see this. A separate PR could change the style.

manifests/client.pp
rvm-peercode: The manpage of my OpenVPN (2.6.14-0ubuntu0.24.04.3) states that the default for data-ciphers is

> AES-256-GCM:AES-128-GCM:CHACHA20-POLY1305 when Chacha20-Poly1305 is available and otherwise AES-256-GCM:AES-128-GCM.

I've used the default here, and the value is only added to the config if changed from the default, so if the user explicitly specifies something else. I though that should ensure it's not a breaking change. Please let me know if you want it differently.

rvm-peercode: BTW It's perfectly fine if you modify this as you see fit, it's been a while since I've worked with this and I don't really have the time to polish it to a large extent I'm afraid.

PR #448, opened 2023-07-03
Add RHEL9

yakatz:
RHEL 9 has the same layout as RHEL 8


yakatz:
Spec failures are due to bad `facterdb` data - fixed by https://github.com/voxpupuli/facterdb/pull/277

PR #452, opened 2024-01-08
Easyrsa version range

jan-win1993:

This pull request addresses 3 problems:
1. Since easy-rsa version 3.0.3 the name of the easy-rsa configuration file did change from "openssl-1.0.cnf" to "openssl-easyrsa.cnf"
The following is a link to the corresponding commit: 
https://github.com/OpenVPN/easy-rsa/commit/e8cd6c980742bb86c34286e655346778275cc36d.
To tackle that problem I would like to introduce easy-rsa version ranges and change the file name if a version above 3.0.3 is used.
2. I did also run into a problem with the "./easy-rsa build-server-full" command which opened an input prompt, which could be suppressed using the "--batch" option
3. The whole openssl output with text gets copied to the client configuration file. I've added a exec resource which runs the "openssl" command with the "-noout" option to suppress the unnecessary text output.

### DIFF
--- a/manifests/ca.pp
+++ b/manifests/ca.pp
@@ -86,8 +86,8 @@ define openvpn::ca (
     require => File["${server_directory}/${name}/easy-rsa"],
   }
 
-  case $openvpn::easyrsa_version {
-    '2.0': {
+  if versioncmp($openvpn::easyrsa_version, '3') == -1 {
+    if versioncmp($openvpn::easyrsa_version, '2') == 1 or versioncmp($openvpn::easyrsa_version, '2') == 0 {
       if $ssl_key_algo != 'rsa' {
         fail('easy-rsa 2.0 supports only rsa keys.')
       }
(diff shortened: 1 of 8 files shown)


fraenki:
@bastelfreak Could we get a review please? Thanks :)

manifests/ca.pp
```diff
@@ -139,13 +139,23 @@
...
+      fail("unexepected value for EasyRSA version, got '${openvpn::easyrsa_version}', expect between 2.0.0 and 3.x.x.")
     }
-    '3.0': {
+  } else {
+    if versioncmp($openvpn::easyrsa_version, '4') == -1 {
+      if versioncmp($openvpn::easyrsa_version, '3.0.3') == 1 {
+        $default_easyrsa_openssl_conf = 'openssl-easyrsa.cnf'
+      } else {
+        $default_easyrsa_openssl_conf = 'openssl-1.0.cnf'
+      }
```
bastelfreak, line 151:
```suggestion
    if versioncmp($openvpn::easyrsa_version, '4') == -1 {
      $default_easyrsa_openssl_conf = if versioncmp($openvpn::easyrsa_version, '3.0.3') == 1 {
        'openssl-easyrsa.cnf'
      } else {
        'openssl-1.0.cnf'
      }
```

bastelfreak:
@fraenki I cannot merge this with the failing jobs. Can you pick up https://github.com/voxpupuli/puppet-openvpn/pull/441? We need to get the latest modulesync in, afterwards we can drop the EoL operating systems, that should get the pipeline green again. I currently don't have the time to work on it.

fraenki:
@jan-win1993 Could you please have a look at this? Rebase on master branch und apply the suggested change. Thanks :)

Valantin:
I think this is useless with the actual master.
Link is disabled and file name changed to the new one.

PR #455, opened 2024-03-12
Drop Debian 9 and 10 support

zilchms:
includes: #454


TheMeier:
Drop Debain 10 too?

zilchms:
ahh right, should drop that too. just focused on the things that make the CI fail :)

TheMeier (approved):

PR #456, opened 2024-03-12
Drop CentOS and Redhat 7 support

zilchms:
includes: #454 #455

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,8 +67,6 @@ jobs:
       fail-fast: false
       matrix:
         setfile:
-          - name: CentOS 7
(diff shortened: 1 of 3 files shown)


TheMeier:
Drop Debian 10 too?

ekohl:
What's the reason to drop CentOS 7? It's not EOL yet. At least, not until June.

zilchms:
Thats actually a good point, I misread the endoflife.org chart...
Time to fix the centos7 failures then

zilchms:
The unit tests are failing due to differences between puppet-facterdb facts for redhat. When those are fixed and released, the unit tests here should be working.
As for the CentOS 7 failures: no idea yet

zilchms:
Dropping CentOS and Redhat 7 as per: https://voxpupuli.org/blog/2024/05/08/CentOS-7-CentOS-Stream-8-deprecation/

bastelfreak (approved):

PR #462, opened 2025-02-05
Allow client_specific_config with extca_enabled

mba-rggir:
There is an issue where the client_specific_config type cannot be used alongside an externally configured Certificate Authority (CA). When extca_enabled is set to true, the existing dependency on openvpn::client causes conflicts, preventing the configuration from working properly. This update eliminates the dependency, enabling the client_specific_config type to function seamlessly with external CA configurations.

This update eliminates the dependency, enabling the client_specific_config type to function seamlessly with external CA configurations.

### DIFF
--- a/manifests/client_specific_config.pp
+++ b/manifests/client_specific_config.pp
@@ -38,7 +38,6 @@ define openvpn::client_specific_config (
 ) {
   if $manage_client_configs {
     Openvpn::Server[$server]
-    -> Openvpn::Client[$name]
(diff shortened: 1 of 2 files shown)


Valantin:
In this commit was introduced the optional you need

https://github.com/voxpupuli/puppet-openvpn/commit/59a6c89bf923200041c9a6a5d09fd5251160b127

Probably documentation and name was not clear but manage_client_config do what you need

PR #469, opened 2025-04-28
Add Ubuntu 24.04 support

Valantin:
Add Ubuntu 24.04 support

#### This Pull Request (PR) include the following PR
#441
#457
#458
#459
#466 
#467
#468


yakatz:
Each PR should only include a single change and shouldn't roll-up other PRs. I will try to figure out which modulesync PRs need to be merged (I think each one is independent, but I am verifying).

Valantin:
> Each PR should only include a single change and shouldn't roll-up other PRs. I will try to figure out which modulesync PRs need to be merged (I think each one is independent, but I am verifying).

Hi, this is a peace of my chain PR

PR #475, opened 2025-05-05
Drop Ubuntu 20.04 support

Valantin:
Drop Ubuntu 20.04 support

### DIFF
--- a/lib/facter/easyrsa.rb
+++ b/lib/facter/easyrsa.rb
@@ -12,7 +12,7 @@ Facter.add(:easyrsa) do
       binaryv3 = '/usr/share/easy-rsa/3/easyrsa'
     when %r{Ubuntu|Debian}
       binaryv3 = case operatingsystemrelease
-                 when %r{|11|12|18.04|20.04|22.04|24.04}
+                 when %r{|11|12|22.04|24.04}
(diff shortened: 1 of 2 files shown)


lib/facter/easyrsa.rb
```diff
@@ -12,7 +12,7 @@
       binaryv3 = '/usr/share/easy-rsa/3/easyrsa'
     when %r{Ubuntu|Debian}
       binaryv3 = case operatingsystemrelease
-                 when %r{|11|12|18.04|20.04|22.04|24.04}
```
bastelfreak, line 15: I think this case statement doesn't make sense anymore? the first when matches all operating system versions?

Valantin (commented):
> I think this case statement doesn't make sense anymore? the first when matches all operating system versions? 

Yes I've a stash with this and other change to remove old code.

kenyon (approved):

lib/facter/easyrsa.rb
TheMeier: also aren't these legacy-facts?

bastelfreak: oh yes, that should also be fixed in line 7/8.

PR #491, opened 2025-09-30
 (draft) Feature/configurable package service

dtapiacl:

### DIFF
--- a/.msync.yml
+++ b/.msync.yml
@@ -2,4 +2,4 @@
 # Managed by modulesync - DO NOT EDIT
 # https://voxpupuli.org/docs/updating-files-managed-with-modulesync/
 
-modulesync_config_version: '10.0.0'
+modulesync_config_version: '10.1.0'
--- a/Gemfile
+++ b/Gemfile
@@ -14,7 +14,7 @@ group :development do
 end
 
 group :system_tests do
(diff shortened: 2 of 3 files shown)


manifests/init.pp
```diff
@@ -51,62 +51,60 @@
...
+
+  # Existing loops unchanged
   $clients.each |$name, $params| {
-    openvpn::client {
-      default:
-        * => $client_defaults;
-      $name:
-        * => $params;
+    openvpn::client { $name:
+      * => merge($client_defaults, $params),
```
bastelfreak, line 89: you don't need to call the merge function, you can use native puppet code:

```suggestion
      * => $client_defaults + $params,
```

manifests/install.pp
```diff
@@ -1,10 +1,21 @@
 #
-# @summary This module installs the openvpn service, configures vpn endpoints, generates client certificates, and generates client config files
+# @summary This module installs and manages OpenVPN (community edition) or OpenVPN Access Server, 
+# configures VPN endpoints, generates client certificates, and generates client config files.
 #
-class openvpn::install {
+# @param package_name
+# The name of the package to install (default: `openvpn`).
```
bastelfreak, line 6: you do not have to mention the default value. puppet-strings will parse the parameter with the datatype and default value, and document all of that in the REFERENCE.md.

